PR #1, opened 2015-08-28
Fix MemoryPathIO and possibility to implement PORT

rsichny:


pohmelie:
Thanks for pull!
Why do we need st_mode? It does not clarify real file permissions for current user. Or I missed something?
I think it's better to rename port to server_port and host to host_port, than breaking straight name forwarding convention (almost straight, cause pass_).
And what about this empty lines before and sometimes after docstrings?

rsichny:
Server expects it to be present in https://github.com/pohmelie/aioftp/blob/master/aioftp/server.py#L875 . I didn't find better solution then just set it to 777 for all files and dirs if using MemoryPathIO

rsichny:
regarding blank lines after docstrings - pep8 ("Method definitions inside a class are surrounded by a single blank line."), before - our internal code style, sorry, will remove those

pohmelie:
Oh, I see... nasty "list"
Anyway it's better to use octal notation 0o100777.
Got it about pep :thumbsup:

rsichny:
@pohmelie changed the PR, check please

pohmelie:
Awesome!

PR #5, opened 2015-09-18
Updated end of line character sequence to CRLF rather than LF

juanqui:
According to RFC 959 (https://www.ietf.org/rfc/rfc959.txt), the end-of-line sequence for FTP commands should be CRLF and not LF. Most FTP server implementations don't care and support either, but Microsoft's IIS/FTP solution is picky and will return a 451 error response due to the invalid line termination (see https://support.microsoft.com/en-us/kb/2505047).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ docs/_build
 /.coverage
 __pycache__
 *.pyc
+*.egg-info
+.idea
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -194,7 +194,7 @@ class BaseClient:
         if command:
 
             common.logger.info(add_prefix(command))
-            self.writer.write(str.encode(command + "\n", encoding="utf-8"))
+            self.writer.write(str.encode(command + "\r\n", encoding="utf-8"))
 
         if expected_codes or wait_codes:
 
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ args = dict(
     name='aioftp',
     version='0.1.8',
     description=('ftp client/server for asyncio'),
-    long_description='\n\n'.join(read('README.rst')),
+    long_description='\n\n'.join(read('readme.rst')),
     classifiers=[
         'License :: OSI Approved :: Apache Software License',
         'Development Status :: 4 - Beta',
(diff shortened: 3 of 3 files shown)


juanqui:
Btw, great project @pohmelie !!

pohmelie:
Awesome!
* Probably we should make constant with name CRLF or EOL in common.py for client and server usage. And what about our mlsd and list workers? Should we send \r\n instead of \n in this case?
* I have no experience with writing setup.py, but can we use version as:
```python
import aioftp
...
args = dict(
    name='aioftp',
    version=aioftp.__version__,
```
* And what about publishing? Cause flit don't collect setup.py to wheel (I think).

Thanks for pull! :+1:

rsichny:
"The argument field consists of a variable length character string
      ending with the character sequence <CRLF> (Carriage Return, Line
      Feed) for NVT-ASCII representation; for other negotiated languages
      a different end of line character might be used"

regarding setup.py - imo it would be better to use configparser and parse flit.ini into setup() args - then we'll have descriptions in one place (flit.ini) though setup.py will be supported too.

rsichny:
I'd wote for client to check "SYST" and in case of "unix" in response use LF, otherwise CRLF.

pohmelie:
@rsichny is there a possibility to server to deal with "\n" and not with "\r\n"? From same rfc:

    End-of-Line
         The end-of-line sequence defines the separation of printing
         lines.  The sequence is Carriage Return, followed by Line Feed.

And if we create duality of LF/CRLF on client, what we should do on server? Server can't determine client system.

rsichny:
@pohmelie guess the server should always use CRLF in command connection, as stated by RFC

pohmelie:
@rsichny yeah, that is what I talking about. So server send CRLF always, but client send CR or CRLF? This looks strange. I mean, if client and server are so «smart», why not use CRLF everywhere?

juanqui:
Technically clients should only send CRLF according to the spec. Some clients only send LF and that's probably a bug or has some legacy reason for being. Servers tend to handle both for maximum compatibility (unless you are Microsoft...)

juanqui:
@pohmelie Didn't mean to add all those commits to this pull request! Just the line ending contribution.

pohmelie:
I will refactor this a bit. Move EOL to common.py and use it both server and client.

PR #9, opened 2015-09-21
fetch setup.py args from flit.ini & add MANIFEST for correct sdist handling

rsichny:


pohmelie:
:+1:

rsichny:
@pohmelie fyi now you should be able to distribute sources as well (generated with `python setup.py sdist`) - this is useful for `pip install -e` for example. Binary distributions are provided by setuptools as well - `python setup.py bdist`, or if you want wheel - `python setup.py bdist_wheel`, but well, you can still use flit.

pohmelie:
@rsichny great! I pretty far away from advanced packaging, that is way I start to use flit.

PR #10, opened 2015-09-21
"python setup.py test" instead of run-tests.py

rsichny:
run tests with `python setup.py test`


rsichny:
hmm, unexpected result on travis, i'll check again...

pohmelie:
https://travis-ci.org/pohmelie/aioftp/builds/81355304
Looks good, what do you mean?

rsichny:
incorrect coverage

pohmelie:
Probably this https://github.com/pohmelie/aioftp/pull/10/files#diff-b873b3ab5a7fcbdf9c691a62c7628153L31 arguments should be here https://github.com/pohmelie/aioftp/pull/10/files#diff-2eeaed663bd0d25b7e608891384b7298R28 too?

rsichny:
nope, they're parsed from setup.cfg

rsichny:
ok, actually `import aioftp` in setup.py breaks it...

pohmelie:
Hm, so, how to deal with __version__?

rsichny:
Maybe move it to separate text file and read it in both __init__.py and setup.py?

There's still problem with coverage - now it does see only one pathio covered...

rsichny:
ok, check it now

pohmelie:
I think it's better to do like "big brother" does https://github.com/KeepSafe/aiohttp/blob/master/setup.py#L47-L53

rsichny:
done.
Actually it was always interesting question - how to handle version in repos...
We do this in our setup.py files, but it won't work correctly with PyPI:
```python
if version is None or release is None:
    try:
        version = subprocess.check_output(['git', '-C', os.path.abspath(os.path.dirname(__file__)), 'describe', '--tags', '--abbrev=0']).decode().strip()
        git_descr = subprocess.check_output(['git', '-C', os.path.abspath(os.path.dirname(__file__)), 'describe', '--tags']).decode().strip()
        if not git_descr.startswith(version):
            raise RuntimeError('Version is "{}" but can not define release from "{}"'.format(version, git_descr))
        release = git_descr[len(version):].lstrip('-').split('-').pop(0)
        if release.isdigit():
            release = str(int(release) + 1)
        elif not release:
            release = '1'
        else:
            raise RuntimeError('Release "{}" that was got from "{}" is not digit...'.format(release, git_descr))
    except (subprocess.CalledProcessError, RuntimeError) as e:
        print(e)
        sys.exit(1)
```

pohmelie:
Yeah, I think someday we will have pep for this :sunglasses:

pohmelie:
Something wrong with logging on error:
```
-------------------- >> begin captured logging << --------------------
'%(asctime)s %(message)s'
'%(asctime)s %(message)s'
'%(asctime)s %(message)s'
'%(asctime)s %(message)s'
'%(asctime)s %(message)s'
--------------------- >> end captured logging << ---------------------
```

rsichny:
oop, sorry, https://github.com/pohmelie/aioftp/pull/11

PR #14, opened 2015-09-22
Provide UserManager so external auth can be implemented

rsichny:
i know that keeping both "users" and "factory" args for server is a bit weird but this way backward compatibility is not broken

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -147,23 +147,56 @@ class User:
 
 
 class UserManager:
+    """
+    Abstract user manager.
+    """
 
     @asyncio.coroutine
     def get_user(self, login):
+        """
+        :py:func:`asyncio.coroutine`
+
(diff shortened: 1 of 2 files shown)


rsichny:
if you're ok with this implementation i'll add docs before you merge

pohmelie:
Great idea of users abstraction layer. The only thing I think is better to change is "user" and "factory". I like the requests api, cause in 99% of cases you don't need all this factories and tune options, so we need to keep it simple. And I think it is easy to do with:
```python
def __init__(self, users=None, *, **):

    if issubclass(users, UserManager):

        self.user_manager = users

    else:
    
        self.user_manager = MemoryUserManager(users)
```
And I will make some style/naming changes later, I think.

rsichny:
done

pohmelie:
Great!

PR #20, opened 2015-09-23
Do not block right after r/w if throttle is used

rsichny:


rsichny:
to explain, your code did this:
```python
while True:
    data = yield from client.read()
    #sleeping because of throttle!!!
    do_something_with_data()
```

after this change we get different behavior:
```python 
while True:
    data = yield from client.read()
    # next operation is not waiting for the sleep time, but we're still throttling
    do_something_with_data()
```

pohmelie:
I don't get why my code should not work? It is asyncio.sleep, just pausing data return… I get your solution, but why mine not working?

rsichny:
it is working, but why pause data return (if it's ready) if we can return it immediately and proceed executing the code?
Consider in example above you will sleep 1 second because of throttle, and do_something_with_data() takes 1 second too. With current throttling implementation one iteration will take 2 seconds, and with lock+call_later - 1 second.

pohmelie:
Oh, I see…

PR #21, opened 2015-09-24
Flake8 integration

rsichny:


rsichny:
an example of error check is here: https://travis-ci.org/pohmelie/aioftp/jobs/81932964

PR #22, opened 2015-09-24
constants should have uppercase names

rsichny:


pohmelie:
I think `END_OF_LINE` is more informative than `CRLF`.

rsichny:
ftplib, smtplib, email.utils - everywhere it's called CRLF ;)

pohmelie:
I know, but this looks wrong for me. What is CRLF? It's just two special characters, why somebody should prefer «strange» abbreviation instead of human name? CRLF about bytes, not about end of line.

rsichny:
changed

pohmelie:
Awesome!

PR #34, opened 2016-01-28
fix st_mode for MemoryPathIO and change build_list_string to use stat…

rsichny:
….filemode

### DIFF
--- a/aioftp/pathio.py
+++ b/aioftp/pathio.py
@@ -4,6 +4,7 @@ import collections
 import operator
 import io
 import time
+import stat
 import sys
 
 from .common import (with_timeout, AsyncStreamIterator, DEFAULT_BLOCK_SIZE,
@@ -41,6 +42,7 @@ class AsyncPathIOContext:
         ... await file.close()
 
     """
+
     def __init__(self, pathio, args, kwargs):
 
         self.close = None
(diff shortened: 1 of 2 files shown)


rsichny:
also changed `lambda` to `def` in asyncpathio to make flake8 happy

aioftp/pathio.py
```diff
@@ -763,19 +766,20 @@ class Lister(AbstractAsyncLister):
         else:
 
             if node.type == "file":
-
```
pohmelie: This empty line should be here. Since whole project written with some code convensions.

aioftp/pathio.py
```diff
@@ -482,7 +484,8 @@ def worker(self):
             @with_timeout
             async def __aiter__(self):
 
-                f = lambda: path.glob("*")
+                def f():
+                    return path.glob("*")
```
pohmelie: Also empty lines before and after `return`

rsichny:
sorry, i'm a bit confused, sometimes you have empty lines in existing code, sometimes you don't... fixed.

pohmelie:
@rsichny 
> sometimes you have empty lines in existing code, sometimes you don't...

Oh, I'm trying at my best :laughing:  If you find such situation please make issue/pull. Code style is important for me.

Pull is great!
The only thing I want to change is `from x import y`, since `stat` is not so long and `stat.y` is a bit cleaner.

Also, I don't remember why I use `append`, but we can just write
``` python
fields = (
    filemode(stats.st_mode),
    ...
)
```
since this is more declarative.

rsichny:
changed

pohmelie:
:+1:

PR #35, opened 2016-02-11
[WIP] Added config to restrict port range for passive server

rsichny:
not for merge yet, tbd:

- [x] change data_ports from list to Queue
- [x] add tests

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -2,6 +2,7 @@ import asyncio
 import pathlib
 import functools
 import datetime
+import errno
 import socket
 import collections
 import enum
@@ -873,6 +874,9 @@ class Server(AbstractServer):
     :param write_speed_limit_per_connection: server write speed limit per
         connection in bytes per second
     :type write_speed_limit_per_connection: :py:class:`int` or :py:class:`None`
+
+    :param data_ports: port numbers that are available for passive connections
+    :type data_ports: :py:class:`collections.Iterable` or :py:class:`None`
     """
     path_facts = (
         ("st_size", "Size"),
@@ -894,7 +898,8 @@ class Server(AbstractServer):
                  read_speed_limit=None,
                  write_speed_limit=None,
                  read_speed_limit_per_connection=None,
-                 write_speed_limit_per_connection=None):
+                 write_speed_limit_per_connection=None,
+                 data_ports=None):
 
         self.loop = loop or asyncio.get_event_loop()
         self.block_size = block_size
@@ -902,6 +907,17 @@ class Server(AbstractServer):
         self.idle_timeout = idle_timeout
         self.wait_future_timeout = wait_future_timeout
         self.path_io = path_io_factory(timeout=path_timeout, loop=self.loop)
+
+        if data_ports is not None:
+
+            self.available_data_ports = asyncio.PriorityQueue(loop=self.loop)
+
+            for data_port in data_ports:
+                self.available_data_ports.put_nowait((0, data_port))
+
+        else:
+
+            self.available_data_ports = None
 
         if isinstance(users, AbstractUserManager):
 
@@ -947,6 +963,7 @@ class Server(AbstractServer):
             client_host=host,
             client_port=port,
             server_host=self.server_host,
+            passive_server_port=0,
             server_port=self.server_port,
             command_connection=stream,
             socket_timeout=self.socket_timeout,
@@ -1041,6 +1058,12 @@ class Server(AbstractServer):
                 if connection.future.passive_server.done():
 
                     connection.passive_server.close()
+
+                    if self.available_data_ports is not None:
+
+                        self.available_data_ports.put_nowait(
+                            (0, connection.passive_server_port)
+                        )
 
                 if connection.future.data_connection.done():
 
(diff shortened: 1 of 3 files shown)


pohmelie:
Good idea to have restricted set of ports.
As I realized, when there is no available port, we create server on random port, and on close connection we add this port to ports pool. So, this mean, that port pool will increase in size with random ports. Is this what user wants?
Another problem is when we use some port from pool and get not "port busy" error, we will not put port back to pool, so pool can became empty.

> change data_ports from list to Queue

You mean collections.deque?

rsichny:
> As I realized, when there is no available port, we create server on random port, and on close connection we add this port to ports pool. So, this mean, that port pool will increase in size with random ports. Is this what user wants?
no, the idea is the following:
1. if ports are not limited - always use port=0 for data connection and don't bother with port pools at all
2. if ports are limited - take first available port from given pool when we start server and put it back when passive server is stopped

>Another problem is when we use some port from pool and get not "port busy" error, we will not put port back to pool, so pool can became empty.
will fix this

>You mean collections.deque?
asyncio.PriorityQueue - it will allow to:
- wait for port if all ports are busy
- lower priority for ports which we couldn't aquire

pohmelie:
> no, the idea is the following

I got the idea, but right now, if port from pool is busy, we use port 0 and put random port back to pool on disconnect.

> wait for port if all ports are busy

How long should we wait? Should this be done with custom timeout? Should we wait at all, or just send error to user? Since passive server socket will exist as long as connection.

rsichny:
updated PR... it works, though:
1. i still don't like the code in _start_passive_server
2. there's a race condition in ```python
should_stop_searching = (
    started_with == connection.passive_server_port
)```

pohmelie:
> i still don't like the code in _start_passive_server

Yeah, but there is place to update some things. I will mark lines since we end with main flow.

> there's a race condition in

Can't realize what kind of race condition you are talking about. Can you explain this?

Maybe move timeout to caller? Since there is couple of await's in `_start_passive_server
`. And I think there should be explicit timeout for this, something like `passive_slot_timeout`, or even more detailed. What do you think?
Also, I think there should be two exceptions: timeout and "no free slot/port".

rsichny:
>Can't realize what kind of race condition you are talking about. Can you explain this?

port which was saved in started_with might be not in Queue:
1. first connection tries to aquire port - it is busy and so just lowered in priority
2. some external app free this port
3. second connection aquires this port while the first connection is still looking for free one in queue
4. port is not in queue anymore and so `should_stop_searching` will never become true

pohmelie:
@rsichny, got it.

What do you think about this solution:
``` python
    async def _start_passive_server(self, connection, handler_callback):

        if self.available_data_ports is not None:

            viewed = set()
            while True:

                port = await self.available_data_ports.get()                                
                try:

                    if port in viewed:

                        raise NoAvailablePort
                        
                    viewed.add(port)
                    passive_server = await asyncio.start_server(
                        handler_callback,
                        connection.server_host,
                        port,
                        loop=connection.loop
                    )
                    connection.passive_server_port = port
                    break

                except:

                    self.available_data_ports.put_nowait(port)
                    _, v, _ = sys.exc_info()
                    if not isinstance(v, OSError):

                        raise

        else:

            passive_server = await asyncio.start_server(
                handler_callback,
                connection.server_host,
                connection.passive_server_port,
                loop=connection.loop,
            )

        return passive_server
```
This should work fine with queue without priority, since order counts for this implementation. Of course we can miss some ports, that will be put back after our first try.

Or, maybe there is still better solution… :confused:

rsichny:
>`if port in viewed:`
i was not sure about keeping a queue of ports just because user might want to have a range of 30k ports for example... and this `in` check might be be slow as hell on high-loaded server ;)

the idea behind priority queue is to lower priority for ports which we couldn't aquire - just to avaoid re-checking them too often

rsichny:
anyway we can try keep that `in`, this should work quite fine (just need to check tests tomorrow):
```
    async def _start_passive_server(self, connection, handler_callback):

        if self.available_data_ports is not None:

            viewed_ports = set()
            while True:

                try:
                    priority, port = self.available_data_ports.get_nowait()

                    if port in viewed_ports:

                        raise NoAvailablePort

                    viewed_ports.add(port)
                    passive_server = await asyncio.start_server(
                        handler_callback,
                        connection.server_host,
                        port,
                        loop=connection.loop
                    )
                    connection.passive_server_port = port
                    break

                except queue.Empty:

                    raise NoAvailablePort

                except (OSError, NoAvailablePort) as err:

                    self.available_data_ports.put_nowait((priority+1, port))
                    if isinstance(err, OSError) and err.errno=errno.EADDRINUSE:
                        pass
                    else:
                        raise

        else:

            passive_server = await asyncio.start_server(
                handler_callback,
                connection.server_host,
                connection.passive_server_port,
                loop=connection.loop,
            )

        return passive_server
```

this also doesn't wait for queue (maybe even better then waiting for port to be released by other connection).

pohmelie:
@rsichny, you are right. Even if we have not so much users and not so much used ports it's bad to collect/check them.
Probably, we should do `qsize()` iterations, instead of `while True`?
``` python
for _ in range(self.available_data_ports.qsize()):
    ...
    break
else:
    raise NoAvailablePort
```
This can produce some overhead via recheck already checked ports, though.

pohmelie:
Also, I see nothing wrong with waiting queue and having explicit timeout for this.

rsichny:
>Also, I see nothing wrong with waiting for queue and having explicit timeout for this.

 i was thinking on it - the chance that other connection will release the port in time is very low - better not to bother with timeouts..

>@rsichny, you are right. Even if we have not so much users and not so much used ports it's bad to collect/check them.
Probably, we should do qsize() iterations, instead of while True?

did some tests today - `in` is not that bad (not really memory-efficient but fast enough /guess because python's `x in set` complexity is O(1)/)

`for` will create more overhead imo

anyway i've updated PR and suppose to accept it's current state for now

aioftp/server.py
```diff
@@ -1042,6 +1059,12 @@ def __init__(self,
 
                     connection.passive_server.close()
 
+                    if self.available_data_ports is not None:
+
+                        self.available_data_ports.put_nowait(
+                            (0, connection.passive_server_port)
+                        )
+
```
pohmelie: If we have no passive connection for session, but have `available_data_ports`, then we will put (0, 0) to queue.

aioftp/server.py
```diff
@@ -1511,12 +1586,18 @@ def get_paths(self, connection, path):
...
+            try:
+
+                connection.passive_server = await coro
+
+            except errors.NoAvailablePort:
+
+                connection.response("421", ["no free ports"])
+
+                return False
```
pohmelie: Should we disconnect user from server in this case?

pohmelie:
> the chance that other connection will release the port in time is very low - better not to bother with timeouts.

True.

> not really memory-efficient

That is my main headache. But I assume, that even if we have much connections, they wont create passive connections simultaneously.

aioftp/server.py
rsichny: as per rfc we should do this:

>421 Service not available, closing control connection.

and there're no other possible replies to PASV except 227, 421 and few 5xx

aioftp/server.py
rsichny: can not reproduce (added `test_data_ports_remains_empty`)

pohmelie: Oh, I miss `if connection.future.passive_server.done()` since this was hidden. My bad.

PR #36, opened 2016-02-18
List should show timestamp for recent files instead of current year

rsichny:


pohmelie:
First of all, when I implement list, I read, that there is no standard for `list` format. So there is nothing good and nothing bad with this pull. So, it will be accepted, since human readability. :ok_hand: 

* What is the reason to change `datetime.datetime.fromtimestamp` to `time.time`?
* Why half of year branching?
* More obvious way to find delta is `timedelta.days`, I think.

rsichny:
it's a copy-paste from asyncssh (https://github.com/ronf/asyncssh/blob/master/asyncssh/sftp.py#L3372), i didn't bother to improve the code - it's simple enough

regarding 6 month it's how \*n\*x `ls -l` works, see http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/ls.c#n4015

PR #50, opened 2017-03-06
 LIST parser and fallback option

jw4js:
LIST parser tested on ftp.gnu.org. Still needs to unescape path names. Inferred the coding style, modify as needed. Take a look at the tests, is there a better way to do it?

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -353,7 +353,7 @@ class BaseClient:
         :return mode:
         :rtype: :py:class:`int`
         """
-        parse_rw = {"rw" : 6, "r-" : 4, "-w" : 2, "--" : 0}
+        parse_rw = {"rw": 6, "r-": 4, "-w": 2, "--": 0}
 
         mode = 0
         mode |= parse_rw[s[0:2]] << 6
@@ -409,10 +409,10 @@ class BaseClient:
 
         :rtype: :py:class:`str`
         """
-        try: 
-           d = time.strptime(s, "%b %d %H:%M")
+        try:
+            d = time.strptime(s, "%b %d %H:%M")
         except ValueError:
-           d = time.strptime(s, "%b %d  %Y")
+            d = time.strptime(s, "%b %d  %Y")
         return time.strftime("%Y%m%d%H%M00", d)
 
     def parse_list_line(self, b):
@@ -698,34 +698,36 @@ class Client(BaseClient):
             >>> stats = await client.list()
         """
         class AsyncListerClient(AsyncListerMixin):
-            
+
             async def _new_stream(cls, local_path):
 
                 cls.path = local_path
+                str_path = " " + str(cls.path)
                 if self.list_compatibility_level > 0:
-                    
+
                     cls._anext = cls._anext_MSLD_LIST
-                    
+
                     if self.list_compatibility_level == 2:
-                    
+
                         cls.process_line = self.parse_mlsx_line
-                    
+
                         try:
-                    
-                            return await self.get_stream(str.strip("MLSD " + str(cls.path)), "1xx")
-                    
+                            command = str.strip("MLSD" + str_path)
+                            return await self.get_stream(command, "1xx")
+
                         except errors.StatusCodeError as e:
-                    
+
                             if e.received_codes[-1] == "500":
-                    
+
                                 self.list_compatibility_level = 1
-                    
+
                     if self.list_compatibility_level == 1:
-                    
+
                         cls.process_line = self.parse_list_line
-                        return await self.get_stream(str.strip("LIST " + str(cls.path)), "1xx")
+                        command = str.strip("LIST" + str_path)
+                        return await self.get_stream(command, "1xx")
                 else:
-                    #TODO: implement NLST/CWD fallback
+                    # TODO: implement NLST/CWD fallback
                     pass
 
             async def __aiter__(cls):
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -807,6 +807,12 @@ class Client(BaseClient):
 
                 return None
 
+        else:
+
+            name, info = self.parse_mlsx_line(str.lstrip(info[1]))
+
+            return info
+
         try:
 
             await self.command("CWD " + str(path.parent), "250")
@@ -825,10 +831,6 @@ class Client(BaseClient):
         finally:
 
             await self.command("CWD " + str(self.cwd), "250")
-
-        name, info = self.parse_mlsx_line(str.lstrip(info[1]))
-
-        return info
 
     async def is_file(self, path):
         """
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -19,6 +19,7 @@ from .common import (
     DEFAULT_BLOCK_SIZE,
     AsyncListerMixin,
     async_enterable,
+    setlocale
 )
 
 __all__ = (
@@ -120,8 +121,8 @@ class BaseClient:
                  read_speed_limit=None,
                  write_speed_limit=None,
                  path_timeout=None,
-                 path_io_factory=pathio.PathIO,
-                 list_compatibility_level=2):
+                 path_io_factory=pathio.PathIO
+                 ):
 
         self.loop = loop or asyncio.get_event_loop()
         self.create_connection = create_connection or \
@@ -139,7 +140,7 @@ class BaseClient:
 
         self.stream = None
         # 2: use MLSD,1:use LIST,0:use NLST/CWD(not implemented)
-        self.list_compatibility_level = list_compatibility_level
+        self.cwd = None
 
     async def connect(self, host, port=DEFAULT_PORT):
 
@@ -158,6 +159,7 @@ class BaseClient:
             timeout=self.socket_timeout,
             loop=self.loop
         )
+        self.list_compatibility_level = 2
 
     def close(self):
         """
@@ -409,10 +411,16 @@ class BaseClient:
 
         :rtype: :py:class:`str`
         """
-        try:
-            d = time.strptime(s, "%b %d %H:%M")
-        except ValueError:
-            d = time.strptime(s, "%b %d  %Y")
+        with setlocale("C"):
+
+            try:
+
+                d = time.strptime(s, "%b %d %H:%M")
+
+            except ValueError:
+
+                d = time.strptime(s, "%b %d  %Y")
+
         return time.strftime("%Y%m%d%H%M00", d)
 
     def parse_list_line(self, b):
@@ -516,6 +524,15 @@ class BaseClient:
 
         return pathlib.PurePosixPath(name), entry
 
+    async def ensure_cwd(self):
+        """
+        Ensures the CWD is a valid value when demanded
+        """
+        if self.cwd is None:
+
+            _, cwd_info = await self.command("PWD", "257")
+            self.cwd = self.parse_directory_response(cwd_info[-1])
+
 
 class Client(BaseClient):
     """
@@ -607,9 +624,12 @@ class Client(BaseClient):
 
         :rtype: :py:class:`pathlib.PurePosixPath`
         """
-        code, info = await self.command("PWD", "257")
-        directory = self.parse_directory_response(info[-1])
-        return directory
+
+        #  code, info = await self.command("PWD", "257")
+        #  directory = self.parse_directory_response(info[-1])
+
+        await self.ensure_cwd()
+        return self.cwd
 
     async def change_directory(self, path=".."):
         """
@@ -620,7 +640,13 @@ class Client(BaseClient):
         :param path: new directory, goes «up» if omitted
         :type path: :py:class:`str` or :py:class:`pathlib.PurePosixPath`
         """
-        if path in ("..", pathlib.PurePosixPath("..")):
+        self.cwd = None
+
+        if type(path) is str:
+
+            path = pathlib.PurePosixPath(path)
+
+        if path == pathlib.PurePosixPath(".."):
 
             cmd = "CDUP"
 
@@ -697,38 +723,32 @@ class Client(BaseClient):
 
             >>> stats = await client.list()
         """
-        class AsyncListerClient(AsyncListerMixin):
+        class AsyncLister(AsyncListerMixin):
 
             async def _new_stream(cls, local_path):
 
                 cls.path = local_path
                 str_path = " " + str(cls.path)
-                if self.list_compatibility_level > 0:
 
-                    cls._anext = cls._anext_MSLD_LIST
+                if self.list_compatibility_level == 2:
 
-                    if self.list_compatibility_level == 2:
+                    cls.parse_line = self.parse_mlsx_line
 
-                        cls.process_line = self.parse_mlsx_line
-
-                        try:
-                            command = str.strip("MLSD" + str_path)
-                            return await self.get_stream(command, "1xx")
-
-                        except errors.StatusCodeError as e:
-
-                            if e.received_codes[-1] == "500":
-
-                                self.list_compatibility_level = 1
-
-                    if self.list_compatibility_level == 1:
-
-                        cls.process_line = self.parse_list_line
-                        command = str.strip("LIST" + str_path)
+                    try:
+                        command = str.strip("MLSD" + str_path)
                         return await self.get_stream(command, "1xx")
-                else:
-                    # TODO: implement NLST/CWD fallback
-                    pass
+
+                    except errors.StatusCodeError as e:
+
+                        if e.received_codes[-1] == "500":
+
+                            self.list_compatibility_level = 1
+
+                if self.list_compatibility_level == 1:
+
+                    cls.parse_line = self.parse_list_line
+                    command = str.strip("LIST" + str_path)
+                    return await self.get_stream(command, "1xx")
 
             async def __aiter__(cls):
 
@@ -737,9 +757,6 @@ class Client(BaseClient):
                 return cls
 
             async def __anext__(cls):
-                return await cls._anext()
-
-            async def _anext_MSLD_LIST(cls):
 
                 while True:
 
@@ -757,7 +774,7 @@ class Client(BaseClient):
 
                             raise StopAsyncIteration
 
-                    name, info = cls.process_line(line)
+                    name, info = cls.parse_line(line)
                     stat = cls.path / name, info
                     if info["type"] == "dir" and recursive:
 
@@ -765,7 +782,7 @@ class Client(BaseClient):
 
                     return stat
 
-        return AsyncListerClient()
+        return AsyncLister()
 
     async def stat(self, path):
         """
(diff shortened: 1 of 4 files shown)


aioftp/client.py
```diff
@@ -120,7 +120,8 @@ def __init__(self,
                  read_speed_limit=None,
                  write_speed_limit=None,
                  path_timeout=None,
-                 path_io_factory=pathio.PathIO):
+                 path_io_factory=pathio.PathIO,
+                 list_compatibility_level=2):
```
pohmelie: I think we don't need "compatibility level", since user should know nothing about `MLSD` or `LIST`, user want to call `client.list()` and that is all. So, implementation should be based on «unknown command» response and fallback to `LIST`.

aioftp/client.py
```diff
@@ -338,6 +341,188 @@ def parse_directory_response(self, s):
...
+        def parse_rw(s):
+            mode = 0
+
+            if s[0] == 'r':
+
+                mode |= 4
+
+            elif s[0] != '-':
+
+                raise ValueError
```
pohmelie: I don't get why server file can't be read-only or write-only?

aioftp/client.py
```diff
@@ -338,6 +341,188 @@ def parse_directory_response(self, s):
...
+
+            return mode
+
+        mode = 0
+        mode |= parse_rw(s[0:2]) << 6
+        mode |= parse_rw(s[3:5]) << 3
+        mode |= parse_rw(s[6:8])
+
+        if s[2] == 's':
```
pohmelie: Maybe some dictionaries instead of imperative way? All bits are just `symbol`-`number` pairs, it looks like it is good ground for declarative description.

aioftp/client.py
```diff
@@ -338,6 +341,188 @@ def parse_directory_response(self, s):
...
+
+            info["type"] = "file"
+
+        elif s[0] == 'd':
+
+            info["type"] = "dir"
+
+        else:
+
+            info["type"] = "unknown"
```
pohmelie: Don't really know what user will do with «unknown»? And what type of object is this?

aioftp/client.py
jw4js: I will fix that

aioftp/client.py
jw4js: In unix, the permissions can be either allowed or disallowed when reading is allowed, it should be 'r', if reading is not allowed, then it should be '-', if reading is not allowed and it is not a '-', then it means something is wrong(bug in our code; bug in the server, or it is an unknown format). If it is read only, it should be 'r--r--r--' if it is write only, then it should be '-w--w--w-', under no (unbuggy) circumstances should it be something other than 'r' or '-' in the read bit, and 'w' or '-' in the write bit

aioftp/client.py
jw4js: See the issue, I'm not sure how the API should expose symlinks

aioftp/client.py
pohmelie: Oh, my bad. It's `!=`.

aioftp/client.py
pohmelie: I've answered in issue thread https://github.com/pohmelie/aioftp/issues/47#issuecomment-284810569

aioftp/client.py
jw4js: Will fix it

aioftp/client.py
```diff
@@ -338,6 +341,188 @@ def parse_directory_response(self, s):
...
+            mode |= 0o0001
+
+        elif s[8] != '-':
+
+            raise ValueError
+
+        return mode
+
+    @staticmethod
+    def parse_ls_date(s):
```
rsichny: this whole method can be rewritten to 4 lines:
```
try: 
   return strptime("%b %e %H:%M", s)
except ValueError:
   return strptime("%b %e  %Y", s)

Comment (jw4js) [inline aioftp/client.py]: The point was to avoid parsing what we don't need to. We have the date in string format, so we just use it as is. But if you insist, I'm fine with using the library functions.

File: aioftp/client.py [hunk 6]
```diff
@@ -338,6 +341,188 @@ def parse_directory_response(self, s):
...
+        :param s: ls date
+        :type s: :py:class:`str`
+
+        :rtype: :py:class:`str`
+        """
+        if not (s[3] == ' ' and s[6] == ' '):
+
+            raise ValueError
+
+        month = "{0:02d}".format(time.strptime(s[:3], "%b").tm_mon)
```
rsichny: the problem here is that this is locale-dependent, i.e. if the locale on server is (for example) ru_RU and on the client en_US - it will just fail. I don't see a good fix for this though.

aioftp/client.py
rsichny: i do not really care just saw a (possibly) cleaner implementation, it's up to @pohmelie

pohmelie: Actually, it is locale dependent. So parsing will break if server and client have different locale.

rsichny: @pohmelie the current implementation is locale-dependent too

pohmelie: I mean month abbreviation.

aioftp/client.py
jw4js: Honestly, there's not much we can do. We can either get a date and time library that supports different locales, or have a monstrous dictionary that has every conceivable locale contained therein.

aioftp/client.py
pohmelie: @rsichny I didn't say opposite :wink: as more as we dig into `LIST`, I understand more and more, that it is «for human only».

pohmelie: @jw4js, probably there is the way to parse date and time with explicit locale…

jw4js: @pohmelie There is, but we don't know what locale it is in before parsing it. So the solution is to just try every locale until one works?

pohmelie:
It looks like parsing/formatting with explicit locale [is possible](http://stackoverflow.com/questions/18593661/how-do-i-strftime-a-date-object-in-a-different-locale), but this should be done pretty accurate.

aioftp/client.py
pohmelie: I think we should stop on English locale, and don't think about other ones. Since `LIST` is complex, but locales make it complex in square.

aioftp/client.py
jw4js: Actually, I left that as an option for easier testing, but I also made an automatic fallback.

jw4js:
@pohmelie Please check new version

aioftp/client.py
```diff
@@ -428,39 +409,11 @@ def parse_ls_date(s):
...
-        if not result.isdigit():
-
-            raise ValueError
-
-        return result + "00"
+        try:
+            d = time.strptime(s, "%b %d %H:%M")
+        except ValueError:
+            d = time.strptime(s, "%b %d  %Y")
+        return time.strftime("%Y%m%d%H%M00", d)
```
pohmelie: We need english locale for non-english users. So, use example from link from my comment https://github.com/pohmelie/aioftp/pull/50#issuecomment-284867736
This context manager should be placed in common, since it will be used in server too.

aioftp/client.py
```diff
@@ -338,6 +341,153 @@ def parse_directory_response(self, s):
...
+
+        # TODO: handle symlinks(beware the symlink loop)
+        info["unix.mode"] = self.parse_unix_mode(s[1:10])
+        s = s[10:].lstrip()
+        i = s.index(' ')
+        info["unix.links"] = s[:i]
+
+        if not info["unix.links"].isdigit():
+
+            raise ValueError
```
pohmelie: It looks like it is enough to just try to convert `s[:i]` to `int`, without next check and raising exception.

aioftp/client.py
pohmelie: This definitely should be removed, since user can disconnect from current server and connect to another, but `list_compatibility_level` already triggered to `LIST`.

aioftp/client.py
```diff
@@ -547,13 +697,38 @@ def list(self, path="", *, recursive=False):
 
             >>> stats = await client.list()
         """
-        class AsyncClientLister(AsyncListerMixin):
+        class AsyncListerClient(AsyncListerMixin):
```
pohmelie: Why name is changed? I'm not native English, but this looks wrong. My naming was like "asynchronous ftp client lister". Maybe `AsyncLister` is enough here.

aioftp/client.py
```diff
@@ -562,6 +737,9 @@ class AsyncClientLister(AsyncListerMixin):
                 return cls
 
             async def __anext__(cls):
+                return await cls._anext()
+
+            async def _anext_MSLD_LIST(cls):
```
pohmelie: Actually in current realization you don't next extra `_anext` method.

aioftp/client.py
```diff
@@ -547,13 +697,38 @@ def list(self, path="", *, recursive=False):
...
+                                self.list_compatibility_level = 1
+
+                    if self.list_compatibility_level == 1:
+
+                        cls.process_line = self.parse_list_line
+                        command = str.strip("LIST" + str_path)
+                        return await self.get_stream(command, "1xx")
+                else:
+                    # TODO: implement NLST/CWD fallback
+                    pass
```
pohmelie: As I said before, you can remove `list_compatibility` level, and just make two blocks: try-except `MLSD`, try-except `LIST` and raise some error if nothing works.

aioftp/client.py
```diff
@@ -547,13 +697,38 @@ def list(self, path="", *, recursive=False):
...
-                command = str.strip("MLSD " + str(cls.path))
-                return await self.get_stream(command, "1xx")
+                str_path = " " + str(cls.path)
+                if self.list_compatibility_level > 0:
+
+                    cls._anext = cls._anext_MSLD_LIST
+
+                    if self.list_compatibility_level == 2:
+
+                        cls.process_line = self.parse_mlsx_line
```
pohmelie: More obvious name for `process_line` is `parse_line` this will echo with `parse_mlsx_line` and `parse_list_line`.

pohmelie:
Also, if `LIST` works fine with files, than maybe it is a good idea to add some `LIST` logic to `stat` function.

aioftp/client.py
jw4js: The problem is when we need to test LIST fallback, we will need a server that doesn't support MLSD. We will need to add an option to the server anyway to not support MLSD if that is to be tested. The alternative is to use ftp.gnu.org, which may support MLSD in the future, and requires the internet to run test cases. We can simply reset `list_compatibility_level` every time we connect to a new server.

aioftp/client.py
jw4js: Will do

aioftp/client.py
jw4js: Ooops, was doing some heavy reformatting and deleted the class, forgot what the name was :(

aioftp/client.py
jw4js: Will fix

aioftp/client.py
jw4js: I think we should cache the result, instead of trying every time we list the directory to see which works.

jw4js:
@pohmelie I actually did add that to stat, but haven't committed it yet. I knew you had more than one thing to say about my current code already so I gave you a break.

aioftp/client.py
jw4js: Well, some weird quirk of Python requires that `__anext__` exists before calling `__aiter__`. There may be a solution to this, but I haven't found it

aioftp/client.py
jw4js: Do you mean to put an int in the info dictionary? I originally decided against it because all other components are strings and to be consistent with MLSD. I also thought to just do `int(s[:i])` on its own line, but decided against that because it is less readable than `isdigit()`. If you want, I can do that, but clarify which one it is that you mean.

aioftp/client.py
pohmelie: Actually you can remove mlsd with 'delattr' inside test.

jw4js: I will do that

aioftp/client.py
jw4js: Actually, I don't think this can be made thread safe, since the calling program might not know about our locale lock. But I will make it anyway.

pohmelie: I think we don't need to make it thread-safe, but make warning in documentation. On other hand, there is [dateparser](https://dateparser.readthedocs.io/en/latest/). Maybe we should use it, I don't really know, since breaking locale is not good at all.

aioftp/client.py
pohmelie: Oh, this is my bad. You are doing it right. Left as is right now.

aioftp/client.py
pohmelie: This is not the place, where we should count performance, I think. Less state is always better. Nothing wrong to call `MLSD` and get `502` here.
> Do you mean to not implement NLST/CWD?
Nope, you can implement it. But will it be usable? Is there ftp server, which have no `MLSD` and no `LIST` support?

aioftp/client.py
pohmelie: I mean, that you don't need to modify `__anext__` at all, since `__anext__` calls `cls._anext`, which is set by `_new_stream` only with one value: `cls._anext_MSLD_LIST`. So you can just change `_anext_MSLD_LIST` to `__anext__`.

aioftp/client.py
jw4js: I originally intended for NLST/CWD to be used in the case we can't parse LIST, but since you decided we don't fallback if we can't parse LIST, then there is no point in implementing that. The caching was not intended for performance, but rather to reduce network traffic, I think we should not bother the server when we don't need to, especially considering how unreliable servers already are(no MLSD). Of course, you have the final decision.

aioftp/client.py
jw4js: Oh, I originally left that to implement NLST/CWD, but since you decided we don't need NLST/CWD, I will remove it.

aioftp/client.py
jw4js: dateparser it is then

aioftp/client.py
jw4js: @pohmelie Please make a decision

jw4js:
Did some caching of the CWD; implemented stat LIST fallback; restructured API as requested; support symlinks in LIST fallback; still needs proper tests(read comment above)

aioftp/client.py
pohmelie: * If the command exists, but we failed on parsing, then we should raise exception, no fallback.
* network traffic is about performance, and this one is pretty cheap. So, no caching right now.

aioftp/client.py
```diff
@@ -620,7 +640,13 @@ class Client(BaseClient):
         :param path: new directory, goes «up» if omitted
         :type path: :py:class:`str` or :py:class:`pathlib.PurePosixPath`
         """
-        if path in ("..", pathlib.PurePosixPath("..")):
+        self.cwd = None
+
+        if type(path) is str:
+
+            path = pathlib.PurePosixPath(path)
```
pohmelie: * For type check `isinstance` is better, since it is the only purpose of `isinstance`.
* You can wrap any input (`str` or `PurePosixPath`) with `PurePosixPath`, so no check is actually needed.

aioftp/client.py
```diff
@@ -603,9 +796,81 @@ class AsyncClientLister(AsyncListerMixin):
...
+
+                else:
+
+                    return None
+
+            else:
+
+                name, info = self.parse_mlsx_line(str.lstrip(info[1]))
+
+                return info
```
pohmelie: You don't need `else` branch here, you can paste `else`-branch code right after `MLST`, since if we got `5xx`, then code below `command` will not be executed.

pohmelie:
@jw4js tests are last part, don't worry about them.

aioftp/client.py
jw4js: :( Didn't know that

aioftp/client.py
jw4js: Fixed

aioftp/client.py
jw4js: Fixed

jw4js:
@pohmelie Tests are the last part, but we are almost finished this issue. Do you mind helping me with the tests, since I don't know how to disable `MLSD`?

pohmelie:
@jw4js, yes, I will make tests.

aioftp/client.py
```diff
@@ -547,13 +722,26 @@ def list(self, path="", *, recursive=False):
...
+
+                cls.parse_line = self.parse_mlsx_line
+
+                try:
+                    command = str.strip("MLSD" + str_path)
+                    return await self.get_stream(command, "1xx")
+
+                except errors.StatusCodeError as e:
+
+                    if e.received_codes[-1] == "500":
```
pohmelie: Use `matches` method of `Code` here. Also, not sure if it is only `500`, since `aioftp.server` sends `502 Command not implemented` on missing command. So, probably `50x` or even `5xx` is more common.

aioftp/client.py
```diff
@@ -547,13 +722,26 @@ def list(self, path="", *, recursive=False):
...
+
+                        cls.parse_line = self.parse_list_line
+                        command = str.strip("LIST" + str_path)
+                        return await self.get_stream(command, "1xx")
```
pohmelie: Fallback part should not be in `except` block. This should be on same level as `try`.

jw4js: Made a fatal mistake in that code, check again. The only way I can think of to put the fallback on the same level as `try` is to set an variable in `except` and add an if after it to test if an exception is caught. I think that's not a good way to do it, but perhaps you were thinking something else?

aioftp/client.py
jw4js: Fixed

aioftp/client.py
```diff
@@ -737,12 +737,14 @@ class AsyncLister(AsyncListerMixin):
...
 
-                    if e.received_codes[-1] == "500":
+                    if e.received_codes[-1].matches("50x"):
 
                         cls.parse_line = self.parse_list_line
                         command = str.strip("LIST" + str_path)
                         return await self.get_stream(command, "1xx")
 
+                    raise
+
```
pohmelie: I mean this logic
``` python
except errors.StatusCodeError as e:

    if not e.received_codes[-1].matches("50x"):

        raise

cls.parse_line = self.parse_list_line
...
```

aioftp/client.py
```diff
@@ -801,12 +803,13 @@ class AsyncLister(AsyncListerMixin):
 
         except errors.StatusCodeError as e:
 
-            if e.received_codes[-1] == "500":
+            if e.received_codes[-1].matches("50x"):
```
pohmelie: Same thing here.

aioftp/client.py
jw4js: :( should have seen that

aioftp/client.py
jw4js: Fixed it

jw4js:
Sorry, didn't mean to generate so many commits, was quite rushed in the morning :(

aioftp/client.py
```diff
@@ -137,6 +139,8 @@ def __init__(self,
         self.path_io = path_io_factory(timeout=path_timeout, loop=loop)
 
         self.stream = None
+        # 2: use MLSD,1:use LIST,0:use NLST/CWD(not implemented)
+        self.cwd = None
```
pohmelie: No need to cash cwd and all this stuff around. Left it as is, since `CWD` is really not using.

aioftp/client.py
```diff
@@ -603,9 +791,46 @@ class AsyncClientLister(AsyncListerMixin):
...
+
+        try:
+
+            code, info = await self.command("MLST " + str(path), "2xx")
+
+        except errors.StatusCodeError as e:
+
+            if not e.received_codes[-1].matches("50x"):
+
+                return None
```
pohmelie: Why return, maybe `raise`?

aioftp/client.py
```diff
@@ -603,9 +791,46 @@ class AsyncClientLister(AsyncListerMixin):
...
+
+        else:
```
pohmelie: No need extra `else` block, on same level as `try`.

aioftp/client.py
```diff
@@ -603,9 +791,46 @@ class AsyncClientLister(AsyncListerMixin):
...
+            path = pathlib.PurePosixPath(path.parts[-1])
+
+            async for file, info in directory_listing:
+
+                if file == path:
+
+                    await directory_listing.stream.finish()
+                    return info
+
+            return None
```
pohmelie: Strange logic, why not straight:
``` python
(_, info), *_ = self.list(path)
return info
```
?

aioftp/client.py
```diff
@@ -603,9 +791,46 @@ class AsyncClientLister(AsyncListerMixin):
...
+
+        finally:
+
+            await self.command("CWD " + str(self.cwd), "250")
```
pohmelie: As above, no need to use `CWD`.

pohmelie:
@jw4js, don't worry, will squash them anyway.

aioftp/client.py
jw4js: Sometimes the file doesn't exist, and below there is that case where LIST doesn't find the file, and there is no exception to raise

aioftp/client.py
jw4js: Didn't know LIST works on a file

aioftp/client.py
jw4js: Fixed

aioftp/client.py
jw4js: Fixed

aioftp/client.py
jw4js: I believe the `else` is needed. Otherwise, it well never reach the fallback

aioftp/client.py
jw4js: Fixed

pohmelie:
@jw4js, thanks for contributing :wink:

PR #54, opened 2017-06-26
server: fix server address in passive mode

yieyu:
when i'm using the default host, the port command did response with 0.0.0.0, thus some clients refusing to connect

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -954,6 +954,7 @@ class Server(AbstractServer):
     async def dispatcher(self, reader, writer):
 
         host, port = writer.transport.get_extra_info("peername", ("", ""))
+        server_host = writer.transport.get_extra_info("sockname")[0]
         message = str.format("new connection from {}:{}", host, port)
         logger.info(message)
 
@@ -972,7 +973,7 @@ class Server(AbstractServer):
         connection = Connection(
             client_host=host,
             client_port=port,
-            server_host=self.server_host,
+            server_host=server_host or self.server_host,
             passive_server_port=0,
             server_port=self.server_port,
             command_connection=stream,
@@ -1626,14 +1627,14 @@ class Server(AbstractServer):
 
             except errors.NoAvailablePort:
 
-                connection.response("421", ["no free ports"])
+                connection.response("421", "no free ports")
                 return False
 
-            code, info = "227", ["listen socket created"]
+            code, info = "227", "listen socket created"
 
         else:
 
-            code, info = "227", ["listen socket already exists"]
+            code, info = "227", "listen socket already exists"
 
         for sock in connection.passive_server.sockets:
 
@@ -1643,7 +1644,7 @@ class Server(AbstractServer):
                 break
 
         nums = tuple(map(int, str.split(host, "."))) + (port >> 8, port & 0xff)
-        info.append(str.format("({})", str.join(",", map(str, nums))))
+        info += str.format(" ({})", str.join(",", map(str, nums)))
 
         if connection.future.data_connection.done():
 


aioftp/server.py
```diff
@@ -972,7 +973,7 @@ def __init__(self,
         connection = Connection(
             client_host=host,
             client_port=port,
-            server_host=self.server_host,
+            server_host=server_host or self.server_host,
```
pohmelie: Wrong order, `self.server_host` should have higher priority I think.

aioftp/server.py
```diff
@@ -1643,7 +1644,7 @@ def get_paths(self, connection, path):
                 break
 
         nums = tuple(map(int, str.split(host, "."))) + (port >> 8, port & 0xff)
-        info.append(str.format("({})", str.join(",", map(str, nums))))
+        info += str.format(" ({})", str.join(",", map(str, nums)))
```
pohmelie: What is the reason for this change?

aioftp/server.py
yieyu: #python3 -m aioftp
aioftp v0.7.0
[21:05:06]: [aioftp.server] serving on 0.0.0.0:2121
[21:05:18]: [aioftp.server] new connection from 192.168.0.192:34212
[21:05:18]: [aioftp.server] 220 welcome
[21:05:18]: [aioftp.server] USER anonymous
[21:05:18]: [aioftp.server] 230 anonymous login
[21:05:18]: [aioftp.server] TYPE I
[21:05:18]: [aioftp.server] 200 
[21:05:18]: [aioftp.server] CWD /
[21:05:18]: [aioftp.server] 250 
[21:05:18]: [aioftp.server] SYST
[21:05:18]: [aioftp.server] 215 UNIX Type: L8
[21:05:18]: [aioftp.server] PASV
[21:05:18]: [aioftp.server] 227-listen socket created
[21:05:18]: [aioftp.server] 227 (0,0,0,0,155,83)

yieyu: here is a failed session with es filemanager client, we notice that the pasv command has **two** repsonse, and the address to be '0,0,0,0'.
this is what pyftpdlib repsonse (a successful session):
[D 2017-06-26 21:08:56] 192.168.0.192:48477-[anonymous] <- PASV
[D 2017-06-26 21:08:56] 192.168.0.192:48477-[anonymous] -> 227 Entering passive mode (192,168,0,110,213,221).

aioftp/server.py
yieyu: a successful pyftpdlib session with only **one** response to PASV command

rsichny: >          Thus the format for multi-line replies is that the first line
>          will begin with the exact required reply code, followed
>          immediately by a Hyphen, "-" (also known as Minus), followed by
>          text.  The last line will begin with the same code, followed
>          immediately by Space <SP>, optionally some text, and the Telnet
>          end-of-line code.

https://tools.ietf.org/html/rfc959

aioftp/server.py
pohmelie: Your answer does not correlate with my question. I'm talking about order of host resolution.
``` python
server=host=self.server_host or writer.transport.get_extra_info("sockname")[0]
```
So, if `server_host` exist, then it should be used.

pohmelie: I see. That is because you start server as `python -m aioftp`, default value for `server_host` is `None` in `server.py`. But in `__main__.py` it is `0.0.0.0`. I think this can be changed to `None` in `__main__.py` and `server_host=self.server_host or writer.transport.get_extra_info("sockname")[0]` logic in `server.py`.

aioftp/server.py
pohmelie: I've just checked `lftp`, and it works fine with multi-line replies. ES have issue, you should write authors about this bug.

pohmelie:
Just realized, that the problem here is not only about `"0.0.0.0"` return to client. When we bind socket with `0.0.0.0` address and we have more than one network interface, then we will be available from any of them. And any user, which will try to make passive connection from "another" network will get wrong address.

aioftp/server.py
legnaleurc: At least this change solves the root cause.
@pohmelie, will you accept this PR if it only contains these two lines?

pohmelie: Yes, definitely. I had just made some tests and it looks like `writer.transport.get_extra_info("sockname")[0]` return proper ip address for different connection sources (`127.0.0.1` and `192.168.1.43` in my case). Anyway, I can add this lines myself if you do not want make this routine.

pohmelie:
I think this can be closed, since

> When we bind socket with 0.0.0.0 address and we have more than one network interface, then we will be available from any of them. And any user, which will try to make passive connection from "another" network will get wrong address.

is wrong, and part of your solution is right. @yieyu, thank you for contribution!

https://github.com/aio-libs/aioftp/commit/a3967472bad8ce1ad255e08bd1bc8725dc5188ce

PR #62, opened 2017-12-29
Handle client side disconnection correctly

jkr78:
If client disconnects during file upload, server exception is not consumed:
https://docs.python.org/3/library/asyncio-dev.html#detect-exceptions-never-consumed

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -886,11 +886,7 @@ class Server(AbstractServer):
                 connection.extra_workers -= done
                 for task in done:
                     try:
-                        try:
-                            result = task.result()
-                        except:  # noqa
-                            logger.exception("dispatcher caught exception")
-                            raise
+                        result = task.result()
                     except errors.PathIOError:
                         connection.response("451", "file system error")
                         continue
@@ -913,8 +909,10 @@ class Server(AbstractServer):
                         else:
                             message = "'{}' not implemented".format(cmd)
                             connection.response("502", message)
-        except ConnectionResetError:
-            logger.info("connection reset by peer")
+        except asyncio.CancelledError:
+            raise
+        except:  # noqa
+            logger.exception("dispatcher caught exception")
         finally:
             message = "closing connection from {}:{}".format(host, port)
             logger.info(message)
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -894,8 +894,6 @@ class Server(AbstractServer):
                     except errors.PathIOError:
                         connection.response("451", "file system error")
                         continue
-                    except ConnectionResetError:
-                        break
                     # this is "command" result
                     if isinstance(result, bool):
                         if not result:
@@ -915,6 +913,8 @@ class Server(AbstractServer):
                         else:
                             message = "'{}' not implemented".format(cmd)
                             connection.response("502", message)
+        except ConnectionResetError:
+            logger.info("connection reset by peer")
         finally:
             message = "closing connection from {}:{}".format(host, port)
             logger.info(message)
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -894,6 +894,8 @@ class Server(AbstractServer):
                     except errors.PathIOError:
                         connection.response("451", "file system error")
                         continue
+                    except ConnectionResetError:
+                        break
                     # this is "command" result
                     if isinstance(result, bool):
                         if not result:


pohmelie:
There is two connections when uploading files: command and data. I don't really now what should happen if one is corrupted. Easy way is to kill all interaction with that user.

Your solution breaks only inner loop, but we are still in outer (`while True`) and actually I'm almost sure, that there can be missed responses from tasks (since you break the loop, in which results are processed) and all flow will be corrupted. Maybe it is better to make decorator for log errors from dispatcher and don't throw them next to asyncio loop, but silently return, since error is logged already?

asvetlov:
> There is two connections when uploading files: command and data. I don't really now what should happen if one is corrupted. Easy way is to kill all interaction with that user.

I support the proposal. On any corruption both sockets should be closed IMHO.

jkr78:
Now should break from both loops.
Tested: simple server `await asyncio.sleep(5 * 60)` during upload and close connection from client. Server side disconnects, on app shutdown - where are no complaints about exceptions never consumed.

pohmelie:
@jkr78 
The problem is that this will handle only one exception. More to say, in your code there is two logger callers for same exception: [here](https://github.com/aio-libs/aioftp/pull/62/files#diff-df09d65446d1f455524ad37a0e1107c5R892) and [here](https://github.com/aio-libs/aioftp/pull/62/files#diff-df09d65446d1f455524ad37a0e1107c5R917)

I think it is better to make logging decorator:
``` python
def log_exception_and_return(*, message, logger=None):

    def decorator(coro):

        @functools.wraps(coro)
        async def wrapper(*args,  **kwargs):
            try:
                return await coro(*args, **kwargs)
            except asyncio.CancelledError:
                raise
            except:  # noqa
                if logger is not None:
                    logger.exception(message)

        return wrapper
        
    return decorator


@log_exception_and_return(message="dispatcher caught exception", logger=logger)
async def dispatcher(self, reader, writer):
    ...    
    try:
        while True:
            done, pending = ...
            for task in done:
                try:
                    result = task.result()                    
                except errors.PathIOError:
                    connection.response("451", "file system error")
                    continue
    ...
```

@asvetlov 
This is default behavior right now, I just wanted some clarification from @jkr78

jkr78:
@pohmelie 

My update was about handling client side disconnection in nicer way. I had no intention to refactor logging. But since we have started discussing - I would do this way:
```python
try:
    while True:
        done, pending = await asyncio.wait(
            pending | connection.extra_workers,
            return_when=asyncio.FIRST_COMPLETED,
            loop=connection.loop
        )
        connection.extra_workers -= done
        for task in done:
            try:
                result = task.result()
            except errors.PathIOError:
                connection.response("451", "file system error")
                continue
            # this is "command" result
            if isinstance(result, bool):
                if not result:
                    await response_queue.join()
                    return
            # this is parse_command result
            elif isinstance(result, tuple):
                pending.add(self.parse_command(stream))
                cmd, rest = result
                if cmd == "pass":
                    # is there a better solution?
                    cmd = "pass_"
                if hasattr(self, cmd):
                    pending.add(getattr(self, cmd)(connection, rest))
                    if cmd not in ("retr", "stor", "appe"):
                        connection.restart_offset = 0
                else:
                    message = "'{}' not implemented".format(cmd)
                    connection.response("502", message)
except ConnectionResetError:
    logger.info("connection reset by peer")
except Exception as e:
    logger.exception("dispatcher caught exception")
finally:
    message = "closing connection from {}:{}".format(host, port)
    logger.info(message)
```
E.g. catching PathIOError's (and other, not fatal, exceptions) as fast as possible and handing them inside loops. Catching fatal exceptions outside the loop. 
I would not reraise exception on unknown error (as in this example). In both ways connection will be closed, but reraising exception makes asyncio library unhappy.

pohmelie:
@jkr78 
Sounds right, looks like decorator is overkill here (until there is another place in library where we need such decorator). I don't understand why `ConnectionResetError` is special? There is no difference with any other critical exception since this is not clean/not regular end of ftp interaction. The only difference for user is that log will have no traceback for `ConnectionResetError`, but have non-informative message. Which connection was reset? What peer it was?

Maybe it is better to handle all exception in one place and add some info about peer (and maybe some debug info).

Also:
* No need to write `except Exception as e:` since you don't use e explicitly, `except:` is enough here.
* Need to reraise `asyncio.CancelledError` (in this case this is not critical maybe, but it is canonical).

jkr78:
@pohmelie 

- agree with `except:` is enought since `e` is not used
- If exception is not reraised, then it can be handled as one `except:`

My point was to catch and do not reraise at least ConnectionResetError. Where is no code who catches this exception afterwards and asyncio complains about not consumed exceptions.

pohmelie:
@jkr78 
I hope this handling is good enough. What do you think?
```python
except asyncio.CancelledError:
    raise
except:
    # some logging on debug and exception levels with peer info (host, port)
```

jkr78:
Yes, looks nice.
Regarding logging, where is already code with logging under `finally`. I think logging exception enough. something like this:
```python
        try:
            while True:
                done, pending = await asyncio.wait(
                    pending | connection.extra_workers,
                    return_when=asyncio.FIRST_COMPLETED,
                    loop=connection.loop
                )
                connection.extra_workers -= done
                for task in done:
                    try:
                        result = task.result()
                    except errors.PathIOError:
                        connection.response("451", "file system error")
                        continue
                    # this is "command" result
                    if isinstance(result, bool):
                        if not result:
                            await response_queue.join()
                            return
                    # this is parse_command result
                    elif isinstance(result, tuple):
                        pending.add(self.parse_command(stream))
                        cmd, rest = result
                        if cmd == "pass":
                            # is there a better solution?
                            cmd = "pass_"
                        if hasattr(self, cmd):
                            pending.add(getattr(self, cmd)(connection, rest))
                            if cmd not in ("retr", "stor", "appe"):
                                connection.restart_offset = 0
                        else:
                            message = "'{}' not implemented".format(cmd)
                            connection.response("502", message)
        except asyncio.CancelledError:
            raise
        except:  # noqa
            logger.exception("dispatcher caught exception")
        finally:
            message = "closing connection from {}:{}".format(host, port)
            logger.info(message)
            if not connection.loop.is_closed():
```

pohmelie:
@jkr78 
Yes, fair enough. Looks good now.

pohmelie:
Thank for your contribution!

PR #77, opened 2018-06-28
fix `__aiter__` must be plain function

Modelmat:
This adds support for Python 3.7 for async generators. That is, 
```py
async def __aiter__(self):
``` 
is not supported in Python 3.7. However, 
```py
def __aiter__(self):
```
is supported in all supported versions of this library.

This also bumps the travis build version.

In instances such as [this](https://github.com/Modelmat/aioftp/blob/3_7_update/aioftp/pathio.py#L452-L476) and [this](https://github.com/Modelmat/aioftp/blob/3_7_update/aioftp/client.py#L601-L679) a workaround has been used to support await statements that were previously in `__aiter__` however this decreases performance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
     - 3.5
     - 3.6
+    - 3.7
 install:
     - travis_retry pip install codecov flake8
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,11 @@ language: python
 python:
     - 3.5
     - 3.6
-    - 3.7
+jobs:
+    include:
+        - python: 3.7
+          dist: xenial
+          sudo: required
 install:
     - travis_retry pip install codecov flake8
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+version: = 0
+conditions: v1
+
 language: python
 python:
     - 3.5
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -628,6 +628,7 @@ class Client(BaseClient):
             >>> stats = await client.list()
         """
         class AsyncLister(AsyncListerMixin):
+            stream = None
 
             async def _new_stream(cls, local_path):
                 cls.path = local_path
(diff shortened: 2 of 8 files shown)


pohmelie:
Oh, travis deactivated `aioftp` repository for some reasons. I have activated it again and fix tests. Update your pull to master, so tests will run without unrelated to this PR errors.

webknjaz:
No it didn't. We moved it to .com. Don't activate .org.

webknjaz:
I've removed travis from services to not confuse you.

Modelmat:
Very confused...?

webknjaz:
See https://github.com/aio-libs/aioftp/pull/77/checks and https://travis-ci.com/aio-libs/aioftp

Also please upgrade badge in README to point to `.com` site.

pohmelie:
> Don't activate .org

I've already did it. Why not?

> We moved it to .com.

Who "we"? Moved what? :confused: 

> I've removed travis from services to not confuse you.

Why so?

asvetlov:
`__aiter__` used to be an async function and became a regular func in Python 3.5.2.

We have two options:
```
        def __aiter__(self):
            return self

        if not PY_352:
            __aiter__ = asyncio.coroutine(__aiter__)
```

and just dropping python < 3.5.2 (or even 3.5.3 as we did in many `aio-libs` projects). 

I vote for the second option.

webknjaz:
I'm doing CI/CD across the whole @aio-libs, so about two months back we moved all repos in this org towards new travis integration. "we" == me and @asvetlov.
Don't enable `.org`, since it's deprecated and sends unnecessary traffic to @travis-ci.

P.S. I'm normally doing announces @ https://github.com/orgs/aio-libs/teams/admins

webknjaz:
> Also please upgrade badge

nvmd I did that myself

webknjaz:
Ref: travis-ci/travis-ci#9815

Modelmat:
Uh I think this was a pull request not a discussion forum... 🤦‍♂️ .

I think we should just drop python 3.5.2 and under .

pohmelie:
@webknjaz 
Oh, got it now.
@asvetlov 
@Modelmat 
I like this idea. Will update `python_requires` in `setup.py` right after this PR will be resolved.
@Modelmat 
It looks like you did not fix all places with `__aiter__`.

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 python:
     - 3.5
     - 3.6
+    - 3.7
```
webknjaz: Please also add it to `allowed_failures:` until travis-ci/travis-ci#9815 resolved. And you may also upvote that request :)

Modelmat: Ok, will do!

Modelmat: Never really done Travis before, is it
```py
matrix:
  allow_failures:
  - python: 3.7
```

pohmelie:
@Modelmat 
Tests failed. I will take a look at this PR localy in couple of hours I hope.

.travis.yml
webknjaz: Nevermind, I'll fix it myself. I've found a way to actually run it in Travis.

webknjaz: but the answer to your question would be yes, that's a correct way to do it.

webknjaz:
I've hooked up a correct config for Python 3.7, but there's one test failing in all envs. @Modelmat feel free to figure out and fix it.

Issue #78, opened 2018-08-29
list() method fails with KeyError when MLSD command returns no facts


thirtyseven:
I am trying to use aioftp as a client for an FTP server with an MLSD command that returns less information than LIST.  Specfically, it returns no facts, which is technically compliant with [RFC 3659](https://tools.ietf.org/html/rfc3659#section-7.2).   This causes the `list()` method to raise a KeyError, as it assumes the response will contain a fact with at least the `type` property:
``` 
File "/Users/tkaplan/.local/share/virtualenvs/ad-pipes-Ecr7n0Zd/lib/python3.7/site-packages/aioftp/client.py", line 672, in __anext__
    if info["type"] == "dir" and recursive:
KeyError: 'type'
```
The fallback to the LIST method is not used in this case, since technically the server responds to the MLSD request.  A nice workaround would be to allow the user to choose whether to use the MLSD or LIST command.

thirtyseven:
Sure, this is the response for an mlsd command in passive mode (actual paths redacted):
```
$ telnet [redacted] 6209
Trying [redacted]...
Connected to [redacted].
Escape character is '^]'.
dir1
dir2
 ..
dir3
dir4
dir5
 .
dir6
dir6
Connection closed by foreign host.
```

pohmelie:
I see three ways to fix this:
* Your suggestion. Add order(?) of commands to try/only one way to try list directory.
* Add more complex fallback logic.
* Both of above. Ability to choose exactly one way to try or try both with complex fallback.

What do you think?

thirtyseven:
I believe it's a proprietary server based on ProFTPD.

PR #79, opened 2018-08-29
Add cmd parameter to list method to allow manually specifying list command.

thirtyseven:
Addresses #78


thirtyseven:
Added some tests

pohmelie:
Thank you for pull and issue! BTW, what is the name of that creepy ftp server, who return no facts?

Issue #37, opened 2016-02-19
Secure FTP


pohmelie:
After #36 I read about FTPS, SFTP and FTP over SSH.
- SFTP is SSH extension and we should ignore it
- FTPS is extension of FTP (https://tools.ietf.org/html/rfc2228)
- FTP over SSH is (as I realized) just tunneling and have problems(?) with data connections

Should be some secure mechanic for FTP implemented?

@rsichny

rsichny:
>FTP over SSH
yes, it's just tunneling...

i was thinking of FTPS - actually this might be quite simple, but as usually the main problem is finding free time for doing things (especially with a full-time job)

i was thinking of FTPS - actually this might be quite simple, but as usually the main problem is finding free time for doing things (especially with a full-time job)

creatorrr:
:+1: Any updates on this? Happy to help out.

pohmelie:
@creatorrr, I tried out [ssl](https://docs.python.org/3/library/ssl.html) module:
``` python
import asyncio
import ssl


req = b"GET / HTTP/1.1\r\nHost: www.python.org\r\n\r\n"


async def flush(outgoing, writer):
    if outgoing.pending:
        writer.write(outgoing.read())
        await writer.drain()


async def foo():
    BLOCK_SIZE = 8192
    # ssl_context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
    ssl_context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
    ssl_context.check_hostname = False

    incoming = ssl.MemoryBIO()
    outgoing = ssl.MemoryBIO()
    ssl_object = ssl_context.wrap_bio(incoming, outgoing)

    reader, writer = await asyncio.open_connection("python.org", 443)
    while True:
        try:
            ssl_object.do_handshake()
            break
        except ssl.SSLWantReadError:
            await flush(outgoing, writer)
            incoming.write(await reader.read(BLOCK_SIZE))

    ssl_object.write(req)
    await flush(outgoing, writer)
    data = None
    while True:
        try:
            data = ssl_object.read(BLOCK_SIZE)
            print(data)
        except ssl.SSLWantReadError:
            await flush(outgoing, writer)

        await flush(outgoing, writer)
        data = await reader.read(BLOCK_SIZE)
        # print("->", data)
        if not data:
            break
        incoming.write(data)

    writer.close()


if __name__ == "__main__":
    loop = asyncio.get_event_loop()
[...]

PR #81, opened 2018-10-09
Add implicit FTPS support

alxpy:
SSL support added.

https://github.com/aio-libs/aioftp/issues/37

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -32,10 +32,11 @@ __all__ = (
 logger = logging.getLogger(__name__)
 
 
-async def open_connection(host, port, loop, create_connection):
+async def open_connection(host, port, loop, create_connection, ssl=None):
     reader = asyncio.StreamReader(loop=loop)
     protocol = asyncio.StreamReaderProtocol(reader, loop=loop)
-    transport, _ = await create_connection(lambda: protocol, host, port)
+    transport, _ = await create_connection(lambda: protocol,
+                                           host, port, ssl=ssl)
     writer = asyncio.StreamWriter(transport, protocol, reader, loop)
     return reader, writer
 
@@ -108,7 +109,8 @@ class BaseClient:
     def __init__(self, *, loop=None, create_connection=None,
                  socket_timeout=None, read_speed_limit=None,
                  write_speed_limit=None, path_timeout=None,
-                 path_io_factory=pathio.PathIO, encoding="utf-8"):
+                 path_io_factory=pathio.PathIO, encoding="utf-8",
+                 ssl=None):
         self.loop = loop or asyncio.get_event_loop()
         self.create_connection = create_connection or \
             self.loop.create_connection
@@ -122,6 +124,7 @@ class BaseClient:
         self.path_io = path_io_factory(timeout=path_timeout, loop=loop)
         self.encoding = encoding
         self.stream = None
+        self.ssl = ssl
 
     async def connect(self, host, port=DEFAULT_PORT):
         self.server_host = host
@@ -131,6 +134,7 @@ class BaseClient:
             port,
             self.loop,
             self.create_connection,
+            self.ssl,
         )
         self.stream = ThrottleStreamIO(
             reader,
@@ -491,6 +495,14 @@ class Client(BaseClient):
 
     :param encoding: encoding to use for convertion strings to bytes
     :type encoding: :py:class:`str`
+
+    :param ssl: if given and not false, a SSL/TLS transport is created
+        (by default a plain TCP transport is created).
+        If ssl is a ssl.SSLContext object, this context is used to create
+        the transport; if ssl is True, a default context returned from
+        ssl.create_default_context() is used.
+        Please look :py:meth:`asyncio.loop.create_connection` docs.
+    :type ssl: :py:class:`bool` or :py:class:`ssl.SSLContext`
     """
     async def connect(self, host, port=DEFAULT_PORT):
         """
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -427,6 +427,7 @@ class AbstractServer:
             host,
             port,
             loop=self.loop,
+            ssl=self.ssl,
             **self._start_server_extra_arguments,
         )
         for sock in self.server.sockets:
@@ -776,6 +777,11 @@ class Server(AbstractServer):
 
     :param encoding: encoding to use for convertion strings to bytes
     :type encoding: :py:class:`str`
+
+    :param ssl: can be set to an :py:class:`ssl.SSLContext` instance
+        to enable TLS over the accepted connections.
+        Please look :py:meth:`asyncio.loop.create_server` docs.
+    :type ssl: :py:class:`ssl.SSLContext`
     """
     path_facts = (
         ("st_size", "Size"),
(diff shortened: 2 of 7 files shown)


asvetlov:
Any chance to add a test?

alxpy:
I thought about tests, but I don't know a good way for testing transport without ssl support at the server part. Can you advise me how to implement it?

pohmelie:
@alxpy, thanks for pr!

This pr is ¼ of what should be implemented for saying «we support ftps». There is two modes: explicit and implicit. And two sides: client and server. You implement implicit client ftps. Also, I'm sceptical about ftps, since it is pretty rare. But, anyway, I see no reasons to not to merge.

@asvetlov 
I'm also want some good tests for this, but as mentioned: without server side implementation tests are not possible in meaning even if they will be implemented they will mock server. I'm not sure that this will be good tests.

@alxpy, is there a chance you will implement implicit server side mode, so we can write tests and have half ftps support?

I will wait a day before merge. Maybe someone have strong point to reject this pr or @alxpy take care of implicit server side support.

asvetlov:
You are right, mocked tests are useless for this case

alxpy:
@pohmelie thanks for the detailed answer!

I will try to implement implicit server side ftps.

rsichny:
lgtm (server side is useles, imo)

aioftp/client.py
```diff
@@ -491,6 +495,13 @@ class Client(BaseClient):
 
     :param encoding: encoding to use for convertion strings to bytes
     :type encoding: :py:class:`str`
+
+    :param ssl: if given and not false, a SSL/TLS transport is created
```
rsichny: please add link to https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection

alxpy: done

alxpy:
Implicit server side FTPS  added.

alxpy:
I can't generate certificate which can work with Python 3.7 =(
https://github.com/aio-libs/aioftp/pull/81/files#diff-572606b9b77c96a18b4bf3eefabc6b25R65
https://bugs.python.org/issue34440

alxpy:
@webknjaz I tried, `trustme` certs doesn't work without `ssl.match_hostname` monkey patching even with Py3.6

webknjaz:
Why? Did you set up the hostname correctly?

tests/common.py
```diff
@@ -42,12 +53,23 @@ def run_in_loop(s_args, s_kwargs, c_args, c_kwargs):
                         client.close()
                     loop.close()
 
+            # SSL monkey patching for tests with self signed certificates.
+            # Works only with Python older than 3.7 version.
+            old_match_hostname = ssl.match_hostname
+            ssl.match_hostname = lambda cert, hostname: True
```
webknjaz: Better use `mock.patch` for such things. It's safer.

tests/test-throttle.py
```diff
@@ -546,15 +546,15 @@ class SlowPathIO(aioftp.PathIO):
 
     async def worker(fname):
 
-        client = aioftp.Client(loop=loop)
-        await client.connect("127.0.0.1", PORT)
-        await client.login()
-        await client.download(
+        _client = aioftp.Client(loop=loop, ssl=client.ssl)
```
webknjaz: There's no point/value in underscoring bunch of vars.

certificates/Certificate.crt
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
webknjaz: Use `trustme` for certs, it's more robust approach.

aioftp/server.py
```diff
@@ -799,7 +805,8 @@ def __init__(self,
                  read_speed_limit_per_connection=None,
                  write_speed_limit_per_connection=None,
                  data_ports=None,
-                 encoding="utf-8"):
+                 encoding="utf-8",
+                 ssl=None):
```
webknjaz: Pro tip: keep closing bracket on a separate line + use trailing commas — this makes diffs nicer.

webknjaz:
@alxpy 
```python
ca = trustme.CA()
ca.issue_server_cert(u'127.0.0.1')
```

alxpy:
Yes, I use hostname correctly. And any tests work, but not all.
I got `ssl.CertificateError: hostname '::1' doesn't match '127.0.0.1'`

aioftp/server.py
alxpy: Personaly I use this code style, but now I use code style of this project...

certificates/Certificate.crt
alxpy: no problem

webknjaz:
oh, that's easy

aioftp/client.py
```diff
@@ -32,10 +32,11 @@
 logger = logging.getLogger(__name__)
 
 
-async def open_connection(host, port, loop, create_connection):
+async def open_connection(host, port, loop, create_connection, ssl=None):
```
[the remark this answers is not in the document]
alxpy: something like `ssl_context` was my first idea, but prefer save original name

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection
https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_server

webknjaz:
@alxpy
```python
ssl_context = ...
...
...
ca = trustme.CA()
ca.issue_server_cert(u'127.0.0.1', u'::1', u'localhost')  # list all incarnations of localhost
ca.configure_trust(ssl_context)
```
Ref: https://trustme.readthedocs.io/en/latest/#trustme.CA.issue_server_cert

aioftp/client.py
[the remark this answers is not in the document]
alxpy: and here `ssl` really can be boolean =)

tests/test-throttle.py
alxpy: so how would you like to call a variable?

webknjaz: keep it as is, I didn't realize that it's an "extended boolean"

alxpy:
`trustme` works fine, thx

pohmelie:
So... @webknjaz @rsichny @asvetlov you think this pr is ready to merge? lgtm. I'm pretty confused, that implicit ftps was so easy to implement.
Hats off to @alxpy :tophat:

tests/test-throttle.py
webknjaz: Oh, that was not about ssl

tests/test-throttle.py
```diff
@@ -293,15 +293,15 @@ class SlowPathIO(aioftp.PathIO):
...
```
webknjaz: Undo underscoring of `client` var. It's not related, nor needed.

webknjaz:
@pohmelie FTPS seems to be just plain old FTP on top of TLS, nothing more. So stick SSL between TCP and the app and you're done. Simple and elegant.
SFTP would be a different story, though, as it works on top of SSH.


@alxpy I insist that wholesale renaming of `client` to `_client` in tests is useless and not necessary. It enlarges the diff and brings no value to tests at all.

pohmelie:
@webknjaz 
I thought at first about explicit ftps: go to secure connection on unsecure command. That sounds pretty hard. Since there must be a method to go down to unsecure from secure. Of course, implicit ftps was pretty straight, but I did not mind, that it is so easy to implement.

alxpy:
>  I insist ...

@webknjaz you can try rename `_client` to `client` and understand the reason for this change.

webknjaz:
@alxpy so tell me: what is that?

pohmelie:
@alxpy Thank you for great contribution! :tada:

PR #84, opened 2018-10-27
Fix deprecation warning in Python 3.7

michalc:
This fixes the warning:

> ... server.py:878: PendingDeprecationWarning: Task.current_task() is
> deprecated, use asyncio.current_task() instead

as documented at https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.current_task:

> This method is deprecated and will be removed in Python 3.9. Use the
> asyncio.current_task() function instead.

The function `asyncio.current_task()` was only added in Python 3.7, and we're supporting earlier, hence the check on Python version.

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -8,6 +8,7 @@ import collections
 import enum
 import logging
 import stat
+import sys
 
 from . import errors
 from . import pathio
@@ -35,6 +36,13 @@ __all__ = (
     "AbstractServer",
     "Server",
 )
+
+IS_PY37_PLUS = sys.version_info[:2] >= (3, 7)
+if IS_PY37_PLUS:
+    get_current_task = asyncio.current_task
+else:
+    get_current_task = asyncio.Task.current_task
+
 logger = logging.getLogger(__name__)
 
 
@@ -875,7 +883,7 @@ class Server(AbstractServer):
             response=lambda *args: response_queue.put_nowait(args),
             acquired=False,
             restart_offset=0,
-            _dispatcher=asyncio.Task.current_task(loop=self.loop),
+            _dispatcher=get_current_task(loop=self.loop),
         )
         connection.path_io = self.path_io_factory(timeout=self.path_timeout,
                                                   loop=self.loop,


.travis.yml
```diff
@@ -3,8 +3,7 @@ conditions: v1
 
 language: python
 python:
-    - 3.5
-    - 3.6
+    - 3.7
```
webknjaz: Please revert this. We need to run tests across a variety of Python versions.

webknjaz:
Hi @michalc,

If you want to contribute, make sure to use shims to make it backward compatible.

.travis.yml
michalc: Have reverted the removal of the tests on 3.5 & 3.6

michalc:
@webknjaz I have added a runtime check on the presence of `asyncio.current_task` ... is that that the sort of shim you mean?

.travis.yml
webknjaz: Now, remove this line, because it's already added under `jobs.include`.

michalc: Ah I misunderstood... done!

aioftp/server.py
```diff
@@ -857,6 +857,11 @@ def __init__(self,
             loop=self.loop
         )
         response_queue = asyncio.Queue(loop=self.loop)
+        if hasattr(asyncio, 'current_task'):
```
webknjaz: Maybe:
```python
IS_PY37 = sys.version_info[:2] == (3, 7)
current_task = asyncio.current_task if IS_PY37 else asyncio.Task.current_task
```

michalc: Changed... (and not sure on the best place for it admittedly)

aioftp/server.py
```diff
@@ -857,6 +858,9 @@ def __init__(self,
             loop=self.loop
         )
         response_queue = asyncio.Queue(loop=self.loop)
+        IS_PY37 = sys.version_info[:2] == (3, 7)
+        current_task = asyncio.current_task if IS_PY37 else asyncio.Task.current_task
+
```
pohmelie: `IS_PY37` should be on top level, to evaluate only once, on import time.

aioftp/server.py
```diff
@@ -857,6 +858,9 @@ def __init__(self,
             loop=self.loop
         )
         response_queue = asyncio.Queue(loop=self.loop)
+        IS_PY37 = sys.version_info[:2] == (3, 7)
+        current_task = asyncio.current_task if IS_PY37 else asyncio.Task.current_task
```
pohmelie: `current_task` should be named as `get_current_task`, I think.

michalc: Done!

aioftp/server.py
michalc: Done

aioftp/server.py
```diff
@@ -857,6 +859,7 @@ def __init__(self,
             loop=self.loop
         )
         response_queue = asyncio.Queue(loop=self.loop)
+        get_current_task = asyncio.current_task if IS_PY37 else asyncio.Task.current_task
```
pohmelie: `get_current_task` also can be moved on top level. And I vote for less foggy "if else". No need to make oneliner.

aioftp/server.py
```diff
@@ -35,6 +36,7 @@
     "AbstractServer",
     "Server",
 )
+IS_PY37 = sys.version_info[:2] == (3, 7)
```
pohmelie: And maybe use `>=` instead of `==`?

pohmelie: I mean rename this as something like `IS_PY37_PLUS`

michalc: Done!

aioftp/server.py
michalc: Done!

pohmelie:
LGTM... @webknjaz any suggestions?
Waiting a day until merge.

PR #85, opened 2018-11-01
Fix Transmit client treating a successful PASV response as failure

michalc:


pohmelie:
I don't get what this solves and why multiline is a problem?

michalc:
@pohmelie I have amended the PR description with more details

pohmelie:
I think at first you should try to get feedback from `Transmit` devs via e-mail (https://library.panic.com/transmit/). Sadly, their **GOLD STANDARD** app have no bugtracker.
```
227-listen socket created
227 (127,0,0,1,154,21)
```
Response is valid and works fine with all tools I tried.

michalc:
I have now emailed them, thanks.

pohmelie:
@michalc any news?

tomharvey:
I'm getting the same error - but not just in Transmit (also seeing an error while using Cyberduck). Also, the PR from @michalc doesn't seem to fix the issue; still `Cannot parse PASV response: listen socket created`.

Maybe this should be an issue instead of in here?

I am using 0.0.0.0 as the bind IP, and then connecting to 127.0.0.1 - is this maybe related to passive not liking the local IP addresses? I can investigate more next weekend..

pohmelie:
> I can investigate more next weekend

It will be good if you have time of course.

PR #86, opened 2018-11-05
Add Windows LIST parser and tests

jw4js:
I don't really like the way the code differentiates between Windows and Unix LIST lines. But I haven't found a better way to implement it with `async`.

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -451,7 +451,7 @@ class BaseClient:
             s = link_src
         return pathlib.PurePosixPath(s), info
 
-    def parse_list_line_windows(self,b):
+    def parse_list_line_windows(self, b):
         """
         Parsing Microsoft Windows `dir` output
 
@@ -468,16 +468,18 @@ class BaseClient:
         # Do not strip! File names can have spaces at the end
         line = s
         i = len(line) - 1
-        # Hopefully files can't have newlines and carriage returns in their name
+        # Hopefully files can't have newlines and
+        # carriage returns in their name
         while line[i] in ['\r', '\n']:
             i -= 1
         line = line[:i + 1]
         date_time_end = line.index('M')
-        date_time_str = line[:date_time_end + 1]
-        date_time_str = ' '.join([x for x in line[:date_time_end + 1].split(' ') if len(x) > 0])
+        date_time_str = line[:date_time_end + 1].strip().split(' ')
+        date_time_str = ' '.join([x for x in date_time_str if len(x) > 0])
         line = line[date_time_end + 1:].lstrip()
         with setlocale("C"):
-            date_time = datetime.datetime.strptime(date_time_str, "%m/%d/%Y %I:%M %p")
+            strptime = datetime.datetime.strptime
+            date_time = strptime(date_time_str, "%m/%d/%Y %I:%M %p")
         info = {}
         info["modify"] = self.format_date_time(date_time)
         next_space = line.index(' ')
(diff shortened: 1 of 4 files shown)


aioftp/client.py
```diff
@@ -687,7 +764,7 @@ def __aiter__(cls):
                         name, info = cls.parse_line(line)
                     except KeyboardInterrupt:
                         raise
-                    except:  # noqa
+                    except Exception as e:  # noqa
```
pohmelie: `# noqa` should be removed in this case. It was old bad times when I don't know about difference between `Exception` and `BaseException`. :smile:

aioftp/client.py
```diff
@@ -440,6 +451,72 @@ def parse_list_line(self, b):
...
+
+        :return: (path, info)
+        :rtype: (:py:class:`pathlib.PurePosixPath`, :py:class:`dict`)
+        """
+        if isinstance(b, bytes):
+            s = b.decode(encoding=self.encoding)
+        else:
+            s = b
+        # Do not strip! File names can have spaces at the end
+        line = s
```
pohmelie: look like you can use `line` above, instead of `s`

aioftp/client.py
```diff
@@ -440,6 +451,72 @@ def parse_list_line(self, b):
...
+        :return: (path, info)
+        :rtype: (:py:class:`pathlib.PurePosixPath`, :py:class:`dict`)
+        """
+        try:
+            return self.parse_list_line_unix(b)
+        except (ValueError, KeyError, IndexError):
+            try:
+                return self.parse_list_line_windows(b)
+            except (ValueError, KeyError, IndexError):
+                raise
```
pohmelie: This should be rewritten in
```
for parser in (unix, windows):
    try:
        ...
```
i think

tests/test-list-fallback.py
```diff
@@ -62,3 +62,38 @@
...
+       6 files           75,978,506,648 bytes
+       3 directories     22,198,362,112 bytes free
+  """
+    test_str = test_str.strip().split("\n")
+    entities = []
+    for x in test_str:
+        try:
+            entities.append(p(aioftp.Client,x))
+        except ValueError as e:
+            pass
```
pohmelie: Don't get this part. Why do you catch `ValueError`? Is it for empty lines? If so filter them with `if`.

tests/test-list-fallback.py
```diff
@@ -62,3 +62,38 @@
...
+    # The spaces in "win7.sh   " are supposed to be there
+    # We parse file names with spaces to the best of our ability
+    correct_names = set(["bin","Desktop","dow","Downloads","msc","opt","win10.img","win10.iso","win10.sh","win7.img","win7.iso","win7.sh   "])
```
pohmelie: Use `{"a", "b", "c"}` for set constructor. And make flake8 happy :smile:

pohmelie:
@toonknapen can you test @jw4js branch in your environment?

tests/test-list-fallback.py
jw4js: Not all lines outputted by `dir` are useful. There is a banner at the beginning and there is some trailing data at the end that is not useful for our purposes. We use `async` and I haven't found a good way to ignore these lines specifically and I'm not sure if the format of those useless lines are the same across versions of `dir`. We also can't be sure if the target server is Unix or Windows until we try to parse a line, meaning we can't simply ignore the first K lines or so on. Currently exceptions are used to indicate that the line can't be parsed.

TL; DR: due to the nature of `async`, `ValueError` is used to ignore the `dir` banners and footer.

jw4js:
I can't seem to get line 512 to pass `flake8`. I broke it into more lines but then `flake8` said it was "under-indented", so I indented more, but now it's "over-indented". :(

tests/test-list-fallback.py
pohmelie: Better approach is `contextlib.suppress` in this case, I think.

pohmelie:
You can use such code to make flake8 happy again:
``` python
parsers = (self.parse_list_line_unix, self.parse_list_line_windows)
for parser in parsers:
```

pohmelie:
Awesome. Last step is to add some info to `history.rst`.

pohmelie:
@jw4js are you stuck on `history.rst` or just have not enough time?

PR #113, opened 2020-06-17
client/list parser: strip date before parsing (fixes #112)

### LINKED ISSUE #112: `parse_ls_date` doesn't handle singe-spaced servers.
I'm using aioftp across a multitude of FTP servers, and just discovered that one of them uses a timestamp of `%b %d %Y` instead of `%b %d  %Y`.
Because the parser assumes the date is exactly 12 characters long, it will catch `Jan 03 2018` as `Jan 03 2018 `. This in turn leads to a datetime parsing error.

ndhansen:
## What do these changes do?

These changes allow the FTP client to parse timestamps that only have a single space between the day and the year.

It does this by stripping the time before parsing it, to eliminate the potential extra space(s) to the right of the time.

No, it simply strips the date string before attempting to parse it.
*However*, it is possible that aioftp is being used on servers that respond with this type of datetime string. These folders will now be picked up, where the in the past would have been ignored.

## Related issue number

#112 

## Checklist


tests/test_simple_functions.py
```diff
@@ -157,6 +157,19 @@ def date_to_p(d):
...
+    def date_to_p(d):
+        return d.strftime("%Y%m%d%H%M00")
+    p = aioftp.Client.parse_ls_date
+    dates = (
+        datetime.datetime(year=2002, month=1, day=1),
+        datetime.datetime(year=2002, month=12, day=31),
+    )
+    for d in dates:
+        s = _c_locale_time(d, format="%b %d %Y")
+        assert p(s) == date_to_p(d)
```
pohmelie, line 170: This test does not cover issue case. `_c_locale_time` will return 11 characters long string.

pohmelie:
> Maybe I'm blind, but I don't see a CHANGES folder.

There is a `history.rst` file.

tests/test_simple_functions.py
ndhansen: Good catch, I've removed the test case as it was useless, and added sample files to the appropriate parsing test below it.

history.rst
```diff
@@ -1,6 +1,12 @@
 x.x.x (xx-xx-xxxx)
 ------------------
 
+0.16.1 (17-06-2020)
```
pohmelie, line 4: Release version and date should be added only on release. Keep changes under `x.x.x`.

ndhansen: :sweat_smile: Thanks for putting up with me, I've just changed it.

PR #114, opened 2020-06-17
client/server: censor the user's password when logging (fixes #111)

### LINKED ISSUE #111: Log output prints password in plain text
I noticed in the log output that aioftp will happily print passwords in the clear, e.g.,

```
INFO:aioftp.client:PASS secret
```

This is very insecure and should be fixed ASAP.

ndhansen:
## What do these changes do?

Currently the user's password is printed in plaintext in the logs. This PR replaces the user's password in log files with stars.

The way this is done was borrowed from https://github.com/python/cpython/blob/master/Lib/ftplib.py#L186

Yes, they will no longer be able to see the password with which they are connecting in the log files

## Related issue number

#111 

## Checklist

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -232,7 +232,11 @@ class BaseClient:
         expected_codes = wrap_with_container(expected_codes)
         wait_codes = wrap_with_container(wait_codes)
         if command:
-            logger.info(command)
+            if command.startswith('PASS'):
+                # Censor the user's password
+                logger.info('PASS ' + '*' * len(command[5:]))
+            else:
+                logger.info(command)
             message = command + END_OF_LINE
             await self.stream.write(message.encode(encoding=self.encoding))
         if expected_codes or wait_codes:
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -243,7 +243,7 @@ class BaseClient:
             if censor_after:
                 # Censor the user's command
                 logger.info(
-                    command[:censor_after] + '*' * len(command[censor_after:])
+                    command[:censor_after] + "*" * len(command[censor_after:])
                 )
             else:
                 logger.info(command)
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -242,9 +242,9 @@ class BaseClient:
         if command:
             if censor_after:
                 # Censor the user's command
-                commands_s = command[:censor_after]
+                raw = command[:censor_after]
                 stars = "*" * len(command[censor_after:])
-                logger.info("%s %s", commands_s, stars)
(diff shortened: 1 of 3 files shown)


pohmelie:
Ok, I think this should be done more gentle. `command` method should receive argument, which describes masking. For example, `self.command("PASV password", ..., shadow_since=5)`.

Also, there is a server part: `parse_command`, which logs all incoming data. In this case there should be a constant list of commands to shadow after parsing.

ndhansen:
Just to make sure I'm understanding you correct, what would `shadow_since` refer to? The index after which to censor the password?

pohmelie:
> The index after which to censor the password?

Yes.

ndhansen:
Perfect, I've added that.

aioftp/server.py
```diff
@@ -481,23 +481,31 @@ async def write_response(self, stream, code, lines="", list=False):
...
         :return: (code, rest)
         :rtype: (:py:class:`str`, :py:class:`str`)
         """
         line = await stream.readline()
         if not line:
             raise ConnectionResetError
         s = line.decode(encoding=self.encoding).rstrip()
-        logger.info(s)
+
+        if s[:4].lower() in censor_commands:
```
pohmelie, line 504: Command directly resolved on 501 line of old version. I think it is better to move this logic with logging below.

aioftp/client.py
```diff
@@ -228,11 +232,21 @@ async def command(self, command=None, expected_codes=(), wait_codes=()):
...
         """
         expected_codes = wrap_with_container(expected_codes)
         wait_codes = wrap_with_container(wait_codes)
         if command:
-            logger.info(command)
+            if censor_after:
+                # Censor the user's command
+                logger.info(
+                    command[:censor_after] + "*" * len(command[censor_after:])
+                )
```
pohmelie, line 247: I think it is better to create "censor"/"non-censor" strings and then concat them without going out of 80 chars.
``` python
command_s = command[:censor_after]
stars = "*" * len(...)
logger.info("%s%s", command_s, stars)
```
or something like this. I'm not a fanat of 80-chars per line right now, but this project historicaly keep this standard.

aioftp/server.py
ndhansen: All right, I've moved it below the splitting thing, and made it more in line with the suggestion for `client.py`.

aioftp/client.py
ndhansen: Yeah, the old line at 78 characters was a bit long :sweat_smile:

I've changed it to reflect your suggestion.

PR #115, opened 2020-06-18
client: allow pass-through of passive commands for directory listing

ndhansen:
## What do these changes do?

When using this library in the wild, I came across an ftp server that times out when attempting to start an `epsv` connection. Because it doesn't return an error code, the client never attempts to switch to regular `pasv`.

This PR allows you to specify the commands with which it will attempt to establish a passive connection to the server in the `list` method itself, and they are then passed down to the `get_passive_connection` method.

Everything is backwards compatible, but it allows the future user to specify which commands to use to establish a passive connection.

## Checklist

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -736,7 +736,9 @@ class Client(BaseClient):
                     try:
                         command = ("MLSD " + str(cls.path)).strip()
                         return await self.get_stream(
-                            command, "1xx", passive_commands=passive_commands
+                            command,
+                            "1xx",
+                            passive_commands=passive_commands
                         )
                     except errors.StatusCodeError as e:
                         code = e.received_codes[-1]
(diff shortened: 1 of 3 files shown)


aioftp/client.py
```diff
@@ -725,15 +735,19 @@ async def _new_stream(cls, local_path):
                 if raw_command in [None, "MLSD"]:
                     try:
                         command = ("MLSD " + str(cls.path)).strip()
-                        return await self.get_stream(command, "1xx")
+                        return await self.get_stream(
+                            command, "1xx", passive_commands=passive_commands
+                        )
```
pohmelie, line 740: For this project it is better to use this style for making "too long calls":
``` python
return await self.get_stream(
    command, 
    "1xx", 
    passive_commands=passive_commands,
)
```

ndhansen: :+1: Fixed it.

pohmelie:
Okay, I think about it and this should be remade, since passive connections created not only for file listing, but for file transfers. This should be moved to `__init__` of `BaseClient` and spread on all passive connections. Maybe create separate PR.

ndhansen:
Closing this in favour of #116

PR #116, opened 2020-07-09
client: add default client-wide passive commands

ndhansen:
## What do these changes do?

When using this library in the wild, I came across an ftp server that times out when attempting to start an epsv connection. Because it doesn't return an error code, the client never attempts to switch to regular pasv.

This change lets you pass default passive commands at creation of the client, and if no custom commands are used when creating the passive connection, these will be used. This means it is now possible for you to set client-wide passive commands which will be used whenever a method calls `get_passive_connection`.

If for some weird reason you were calling `get_passive_connection` with an empty list of commands, it will no longer result in a `ValueError`, because the default client passive commands will be used. But I have no idea why anyone would *ever* do that.

## Checklist

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -1098,7 +1098,6 @@ class Client(BaseClient):
             commands = self._passive_commands
         if not commands:
             raise ValueError("No passive commands provided")
-
         await self.command("TYPE " + conn_type, "200")
         for i, name in enumerate(commands, start=1):
             name = name.lower()
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -109,7 +109,7 @@ class BaseClient:
                  read_speed_limit=None, write_speed_limit=None,
                  path_timeout=None, path_io_factory=pathio.PathIO,
                  encoding="utf-8", ssl=None, parse_list_line_custom=None,
-                 **siosocks_asyncio_kwargs):
+                 passive_commands=("epsv", "pasv"), **siosocks_asyncio_kwargs):
         self.socket_timeout = socket_timeout
         self.throttle = StreamThrottle.from_limits(
             read_speed_limit,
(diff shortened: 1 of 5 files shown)


ndhansen:
@pohmelie do you think we could put out a patch to pypi when this PR is done?

aioftp/client.py
```diff
@@ -1092,18 +1094,24 @@ async def get_passive_connection(self, conn_type="I",
...
             "pasv": self._do_pasv,
         }
-        if not commands:
+        if not commands and not self._passive_commands:
             raise ValueError("No passive commands provided")
+
+        passive_commands = self._passive_commands
+        if commands is not None:
+            # Overwrite the default passive commands
+            passive_commands = commands
```
pohmelie, line 1103: Strange logic. I think it is better to write something like
``` python
if not commands:
    commands = self._passive_commands
if not commands:
    raise ValueError(...)
```

tests/test_passive.py
```diff
@@ -20,11 +20,29 @@ async def test_client_fail_fallback_to_pasv_at_list(pair_factory,
...
+async def test_client_only_enhanced_passive_list(pair_factory):
+    async with pair_factory(host="127.0.0.1") as pair:
+        pair.client._passive_commands = ["epsv"]
+        await pair.client.list()
 
 
 @pytest.mark.asyncio
 async def test_passive_no_choices(pair_factory):
     async with pair_factory() as pair:
+        pair.client._passive_commands = tuple()
```
pohmelie, line 45:
```suggestion
        pair.client._passive_commands = []
```

aioftp/client.py
ndhansen: Good point. I've changed it.

tests/test_passive.py
ndhansen: :+1: Fixed.

pohmelie:
> @pohmelie do you think we could put out a patch to pypi when this PR is done?

Yes.

aioftp/client.py
```diff
@@ -1092,8 +1094,11 @@ async def get_passive_connection(self, conn_type="I",
             "epsv": self._do_epsv,
             "pasv": self._do_pasv,
         }
+        if not commands:
+            commands = self._passive_commands
         if not commands:
             raise ValueError("No passive commands provided")
+
```
pohmelie, line 1101:
```suggestion
```

PR #122, opened 2020-08-01
docs: fix extra parentheses

Olegt0rr:
Extra parentheses fix


pohmelie:
Nice catch! This should be changed in `docs/index.rst` too.

PR #125, opened 2020-08-24
Fix Pasv Listening On 0.0.0.0, Fix MLSD Not Show

PonyPC:
Fix Pasv Listening On 0.0.0.0

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -419,8 +419,6 @@ class AbstractServer(abc.ABC):
                 host, port, *_ = sock.getsockname()
                 if not self.server_port:
                     self.server_port = port
-                if not self.server_host:
-                    self.server_host = host
                 logger.info("serving on %s:%s", host, port)
 
     async def serve_forever(self):
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -1141,6 +1141,7 @@ class Server(AbstractServer):
         raw = await connection.path_io.stat(path)
         for attr, fact in Server.path_facts:
             stats[fact] = getattr(raw, attr)
+        stats['Modify'] = time.strftime('%Y%m%d%H%M%S', time.gmtime(stats['Modify']))
         s = ""
         for fact, value in stats.items():
             s += f"{fact}={value};"


pohmelie:
Can you describe the problem and provide snippet to reproduce. I did not get the reason for changes.

PonyPC:
Start the server, and filezilla connect to.
List and mlsd failed, because pasv return 0,0,0,0,xx,xx.
Also mlsd modify time is incorrect so that the client cannot show file list.
Both are right in version 0.12.0

pohmelie:
1. How did you start the server? `lftp` works fine with `python -m aioftp --family ipv4` and zeros in `PASV` response. When you create listen socket without describing hostname/ip address, socket will be created on all interfaces, and `getsockname` returns `0.0.0.0` in this case. `0.0.0.0` is «de facto» standard for many protocols to say «use same host you use at current connection». I got the problem, but it is not a problem in common case. I see three solutions:
- you can start server with exact interface (e.g. `python -m aioftp --host 123.456.78.9`)
- ipv4 `0.0.0.0` should be ignored when discovering binded hostname
- there should be an option for «which address to choose: binded or connected to» 
2. mlsx response have no changes since 0.12.0, anyway, it looks like it is broken, since it's really should be in another format.

This two problems should be resolved in two separate PR's because first is discussable, second is obvious and pretty simple. And it is my bad that time is in wrong format right now.

PonyPC:
I tried family with socket.AF_INET4, address 0.0.0.0, still has this problem. In version 0.12.0, PASV return use same host you use at current connection. Mostly we set bind address to None for adapting automatically:
```
aioftp.Server().start(None, 21, reuse_port=True)
```
But in version lasted, PASV return 0,0,0,0. 
I prefer choose second solution like version 0.12.0

pohmelie:
Agreed, this should be optionated. I will make an option for this in couple of days, since afk right now.

PonyPC:
Thanks.

pohmelie:
I've just pushed changes on `PASV` server host resolving. No extra options, address will be taken from connection since now. Since I did not found any case where address should be taken from binded sockets. New version `0.18.0` released on pypi. Feel free to create issue if something went wrong. Will close this PR.

PR #131, opened 2020-10-02
log debug-like messages at debug level

ch3pjw:
## What do these changes do?

Make logging output less noisy when used in client applications

Yes - if the user sets up logging they will see fewer log messages at `info`, which in my case will help make application log messages easier to spot.

## Checklist


ch3pjw:
Thanks :+1: When might this be released to PyPi?

pohmelie:
Released just now.

PR #132, opened 2020-10-05
Ability to set TCP keepalive to detect/closed dead connections

eLvErDe:
Hello,

I had trouble on a server having tones of dead connection to remote FTP server in CLOSED_WAIT state. I'm not sure what exactly caused that (I suspect the remote server had issues) but still, TCP keepalive mecanism is here to avoid this.
So here's a new kwarg to aioftp Client providing the ability to request TCP keepalive.

Regards, Adam.


pohmelie:
I see a better solutions here, make `socket` property for client, which will be available after connection made. So anyone can tune it as it wish.

eLvErDe:
> I see a better solutions here, make `socket` property for client, which will be available after connection made. So anyone can tune it as it wish.

Except when using as context manager, this is why I did this that way

pohmelie:
You always can use your own context manager with socket tune after `connect` call. Current approach lacks of flexibility and is to strict. Some time later someone will want another socket option to tune.

eLvErDe:
> You always can use your own context manager with socket tune after `connect` call. Current approach lacks of flexibility and is to strict. Some time later someone will want another socket option to tune.

Well it's your call anyway. For me these options are quite important so I like the idea of having them exposed as parameters with proper documentation

pohmelie:
Sorry for a delay. I've just released, that `stream` is a public client attribute. So you can use something like:
``` python
sock = client.stream.writer.get_extra_info("socket")
...
```
And I think there should not be second network connection abstraction like `socket`.

PR #137, opened 2021-01-27
server/mlsx: add path existance check to prevent symlink `stat` call failure

PonyPC:
Fix MLSD error on listing corrupt symbol link file

Some corrupt symbol link file point to non-exists file will cause MLSD error.

## Checklist

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -1134,9 +1134,10 @@ class Server:
 
     async def build_mlsx_string(self, connection, path):
         if not await connection.path_io.exists(path):
-            return ""
-        stats = await connection.path_io.stat(path)
-        facts = self._build_mlsx_facts_from_stats(stats)
+            facts = {}
+        else:
+            stats = await connection.path_io.stat(path)
+            facts = self._build_mlsx_facts_from_stats(stats)
         if await connection.path_io.is_file(path):
             facts["Type"] = "file"
         elif await connection.path_io.is_dir(path):
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -1133,7 +1133,10 @@ class Server:
         }
 
     async def build_mlsx_string(self, connection, path):
-        stats = await connection.path_io.stat(path)
+        try:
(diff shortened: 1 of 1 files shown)


pohmelie:
I'm not sure that it is a timeout, since on path exception raised in MLSD task there will be a proper response with 451 code (https://github.com/aio-libs/aioftp/blob/master/aioftp/server.py#L938)

PonyPC:
You're right. It won't list directory instead timeout.

aioftp/server.py
```diff
@@ -1133,7 +1133,10 @@ def _build_mlsx_facts_from_stats(self, stats):
         }
 
     async def build_mlsx_string(self, connection, path):
-        stats = await connection.path_io.stat(path)
+        try:
+            stats = await connection.path_io.stat(path)
+        except asyncio.exceptions.InvalidStateError:
```
pohmelie, line 1138: Almost all exceptions in path methods will be reraised as [`PathIOError`](https://github.com/aio-libs/aioftp/blob/master/aioftp/pathio.py#L88). I suggest to not use exception as flow changer. I think it is better to check that path exists, before calling stat.

PonyPC: Okay, I have modified by your direction

aioftp/server.py
```diff
@@ -1133,6 +1133,8 @@ def _build_mlsx_facts_from_stats(self, stats):
         }
 
     async def build_mlsx_string(self, connection, path):
+        if not await connection.path_io.exists(path):
+            return ""
         stats = await connection.path_io.stat(path)
         facts = self._build_mlsx_facts_from_stats(stats)
```
pohmelie, line 1139: I think empty spec is a bad idea. Client will get empty line and it is pretty confused. At least Type=unknown and filename will be in spec line with such changes
```suggestion
        if not await connection.path_io.exists(path):
            facts = {}
        else:
            stats = await connection.path_io.stat(path)
            facts = self._build_mlsx_facts_from_stats(stats)
```

PR #148, opened 2022-01-27
server: fix real directory resolve for windows (fixes #147)

### LINKED ISSUE #147: Directory traversal / base path escape possible on Windows
The server allows escaping the user `base_path` on Windows using absolute path (segments).
This is due to [`get_paths`](https://github.com/aio-libs/aioftp/blob/81d4bc018261268bf2ea88f42e05f130ab4c9229/aioftp/server.py#L991):
While the virtual path handling is perfectly fine in itself (it represents a POSIX path, so all is good there), the joining between `base_path` and `resolved_virtual_path` is not and quickly falls apart when Windows absolute paths are involved.
In my opinion this presets a serious security issue when developers naively rely on the `base_path` setting to not allow users to traverse outside of it.

Consider the following:
```python
base_path = pathlib.Path("C:\\locked_path")  # base_path is of the type Path (so platform dependent path)
```
Now see what happens when the input virtual path contains Windows absolute path segments:
```python
resolved_virtual_path = pathlib.PurePosixPath("/foo/C:\\Windows/bla")
real_path = base_path / resolved_virtual_path.relative_to("/")
>>> PureWindowsPath('C:/Windows/bla')
```
You have now broken out.
This is due to two reasons:
1. While joining, `pathlib` joins each path segment to the list of first path object with the following rule:
    > When several absolute paths are given, the last is taken as an anchor (mimicking os.path.join()’s behaviour)
    This is also the reason the `relative_to("/")` does not help here.
2. When joining, `pathlib` joins according to the rules of the first path object, so in this case it pulls out all path `parts` from the `PurePosixPath` and joins them to the `PureWindowsPath` by the rules of `PureWindowsPath`, meaning the Windows absolute anchor in the `PurePosixPath` is interpreted as on Windows.

There are a myriad of ways to _almost_ fix it inside the virtual path, but they always have special cases that break them and I do not consider them safe.
In my opinion, the best way to prevent any directory traversal using `pathlib` is by going by the resulting absolute path.
1. Join the `resolved_virtual_path` to the `base_path` as is already done, but using the anchor given in the path to be relative to since users can pass Windows anchors too:
   ```
   real_path = base_path / resolved_virtual_path.relative_to(resolved_virtual_path.anchor)
   ```
2. Use `relative_to` to make sure the joined `real_path` is still relative to / inside the `base_path` and give us the fully resolved (all absolute anchors removed) subdirectory/virtual path at the same time:
   ```
   resolved_virtual_path = pathlib.PurePosixPath("/") / real_path.relative_to(base_path)
   ```
3. If the previous call raises (with our example from above: `ValueError: 'C:\\Windows\\bla' is not in the subpath of 'C:\\locked_path' OR one path is relative and the other is absolute.`) we know that a traversal was attempted and can either return 
    1. An access denied error (not very fitting in the current design, would need to either convert to using exceptions or add lots of boilerplate)
    2. Just return something like `base_path, pathlib.PurePosixPath("/")` to throw the user back to the home directory.

I will submit a pull request shortly.

ported-pw:
## What do these changes do?

Sanitize paths so directory traversal is not possible anymore.

Only for users trying to exploit directory traversal. It will send these back to the home directory.

## Related issue number

https://github.com/aio-libs/aioftp/issues/147

## Checklist

### DIFF
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -988,7 +988,8 @@ class Server:
             if tasks_to_wait:
                 await asyncio.wait(tasks_to_wait)
 
-    def get_paths(self, connection, path):
+    @staticmethod
+    def get_paths(connection, path):
         """
         Return *real* and *virtual* paths, resolves ".." with "up" action.
         *Real* path is path for path_io, when *virtual* deals with
@@ -1013,11 +1014,13 @@ class Server:
             else:
                 resolved_virtual_path /= part
         base_path = connection.user.base_path
-        real_path = base_path / resolved_virtual_path.relative_to(resolved_virtual_path.anchor)
+        real_path = base_path / resolved_virtual_path.relative_to("/")
+        # replace with `is_relative_to` check after 3.9+ requirements lands
         try:
-            resolved_virtual_path = pathlib.PurePosixPath("/") / real_path.relative_to(base_path)
+            real_path.relative_to(base_path)
         except ValueError:
-            return base_path, pathlib.PurePosixPath("/")
+            real_path = base_path
+            resolved_virtual_path = pathlib.PurePosixPath("/")
         return real_path, resolved_virtual_path
 
     async def greeting(self, connection, rest):
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -1013,7 +1013,11 @@ class Server:
             else:
                 resolved_virtual_path /= part
         base_path = connection.user.base_path
-        real_path = base_path / resolved_virtual_path.relative_to("/")
+        real_path = base_path / resolved_virtual_path.relative_to(resolved_virtual_path.anchor)
+        try:
+            resolved_virtual_path = pathlib.PurePosixPath("/") / real_path.relative_to(base_path)
+        except ValueError:
+            return base_path, pathlib.PurePosixPath("/")
         return real_path, resolved_virtual_path
 
     async def greeting(self, connection, rest):
--- a/tests/test_simple_functions.py
+++ b/tests/test_simple_functions.py
@@ -230,3 +230,17 @@ def test_server_mtime_build():
     b = aioftp.Server.build_list_mtime
     assert b(now, now) == "Jan  1 00:00"
     assert b(past, now) == "Jan  1  2001"
+
+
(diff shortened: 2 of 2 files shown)


aioftp/server.py
```diff
@@ -1013,7 +1013,11 @@ def get_paths(self, connection, path):
             else:
                 resolved_virtual_path /= part
         base_path = connection.user.base_path
-        real_path = base_path / resolved_virtual_path.relative_to("/")
+        real_path = base_path / resolved_virtual_path.relative_to(resolved_virtual_path.anchor)
+        try:
+            resolved_virtual_path = pathlib.PurePosixPath("/") / real_path.relative_to(base_path)
+        except ValueError:
+            return base_path, pathlib.PurePosixPath("/")
```
pohmelie, line 1020:
```suggestion
        real_path = base_path / resolved_virtual_path.relative_to("/")
        try:
            real_path.relative_to(base_path)
        except ValueError:
            real_path, resolved_virtual_path = base_path, pathlib.PurePosixPath("/")
```

ported-pw: 1. I chose to replace `"/"` with `.anchor` mostly because in itself it's the usual pattern to remove the anchor. It's not technically necessary here as you pointed out but in my opinion it makes the line in itself more correct (and doesn't really make anything worse either apart from making the line slightly longer)
2. Why remove re-creating the virtual path? I suppose it's just a cosmetic issue since it doesn't affect the `real_path` (which is the important part) in any way but doing it the way I did ensures there's nothing else going on there that wouldn't lead to a traversal outside of `base_path`

pohmelie: 1. I think `"/"` is more explicit and readable.
2. There is no need to recreate virtual path, since it is created already and can be «valid» or «invalid». Check that real path is in bounds of base path is enough.

ported-pw: Both points make sense to me, feel free to go ahead and implement your version, however I can't change it in the PR branch for now since that would require us re-testing the changed behaviour.

pohmelie:
@ported-pw 
Thank you for the contribution.

PR #165, opened 2023-07-03
add type hints

honglei:
tested under Debian10/python3.7, Debian10/python3.11,

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -6,11 +6,11 @@ import datetime
 import logging
 import pathlib
 import re
-from collections.abc import Awaitable, Callable
+from collections.abc import AsyncIterable, Awaitable, Callable
 from functools import partial
 from ssl import SSLContext
-from typing import Literal, Any, Iterator, Generator, TypeVar, Type, AsyncIterable, Self
 from types import TracebackType
+from typing import Any, Literal, Self, TypeVar
 
 from aioftp import errors, pathio
 from aioftp.common import (
@@ -1259,7 +1259,7 @@ class Client(BaseClient):
     @classmethod
     @contextlib.asynccontextmanager
     async def context(
-        cls: Type[T],
+        cls: type[T],
         host: str,
         port: int = DEFAULT_PORT,
         user: str = DEFAULT_USER,
--- a/aioftp/common.py
+++ b/aioftp/common.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
+
 import abc
 import asyncio
 import collections
 import functools
 import locale
 import threading
+from collections.abc import Awaitable, Callable, Iterator
 from contextlib import contextmanager
-from typing import Any, Self, Callable, Awaitable, Iterator
 from types import TracebackType
+from typing import Any, Self
 
 __all__ = (
     "AbstractAsyncLister",
--- a/aioftp/errors.py
+++ b/aioftp/errors.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
-from typing import Any
+
 from types import TracebackType
+from typing import TYPE_CHECKING, Any
+
 from aioftp import common
-from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from aioftp.client import Code
--- a/aioftp/pathio.py
+++ b/aioftp/pathio.py
@@ -9,8 +9,9 @@ import pathlib
 import stat
 import sys
 import time
-from typing import Self, BinaryIO, IO, Any, List
 from io import BufferedRandom
+from typing import Any, BinaryIO, List, Self
+
 from aioftp import errors
 from aioftp.common import (
     DEFAULT_BLOCK_SIZE,
@@ -50,7 +51,7 @@ class AsyncPathIOContext:
 
     """
 
-    def __init__(self, pathio: "AbstractPathIO", args: List[Any], kwargs: dict[Any, Any]):
+    def __init__(self, pathio: "AbstractPathIO", args: list[Any], kwargs: dict[Any, Any]):
         self.close = None
         self.pathio = pathio
         self.args = args
--- a/aioftp/server.py
+++ b/aioftp/server.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 import abc
 import asyncio
 import enum
@@ -10,8 +11,8 @@ import pathlib
 import socket
 import stat
 import time
-from typing import Any, Awaitable, Self
-from collections.abc import Callable, Iterator, Sequence
+from collections.abc import Awaitable, Callable, Iterator, Sequence
+from typing import Any, Self
 
 from aioftp import errors, pathio
 from aioftp.common import (
@@ -391,9 +392,9 @@ class ConnectionConditions:
         self.fail_code: str = fail_code
         self.fail_info: str | None = fail_info
 
-    def __call__(self, f: Callable[["Server", Connection, str, list[Any]], Awaitable[bool]]) -> Any:
+    def __call__(self, f: Callable[[Server, Connection, str, list[Any]], Awaitable[bool]]) -> Any:
         @functools.wraps(f)
-        async def wrapper(cls: "Server", connection: Connection, rest: str, *args: list[Any]) -> bool:
+        async def wrapper(cls: Server, connection: Connection, rest: str, *args: list[Any]) -> bool:
             futures = {connection[name]: msg for name, msg in self.fields}
             aggregate = asyncio.gather(*futures)
             if self.wait:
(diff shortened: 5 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
         with:
           fail_ci_if_error: true
           verbose: true
+          token: ${{ secrets.CODECOV_TOKEN }}
```
pohmelie, line 35: Why we need this?

aioftp/client.py
```diff
@@ -7,6 +7,8 @@
 import pathlib
 import re
 from functools import partial
+from typing import Optional, Union, Type, Tuple, Dict, List
+# from collections.abc import Sequence, Iterator
```
pohmelie, line 11: This comment is a mistake, I think

aioftp/client.py
```diff
@@ -7,6 +7,8 @@
 import pathlib
 import re
 from functools import partial
+from typing import Optional, Union, Type, Tuple, Dict, List
```
pohmelie, line 10: We can target python 3.11, I think for better typing style like `x: int | float | None = None`

aioftp/client.py
```diff
@@ -45,7 +47,8 @@ class Code(str):
     """
     Representation of server status code.
     """
-    def matches(self, mask):
+
+    def matches(self, mask: str):
```
pohmelie, line 51: We should add return types of all functions, to have strict mypy checks

aioftp/client.py
```diff
@@ -75,6 +78,7 @@ class DataConnectionThrottleStreamIO(ThrottleStreamIO):
     :param **kwargs: keyword arguments passed to
         :py:class:`aioftp.ThrottleStreamIO`
     """
+
```
pohmelie, line 81: Why this empty string here?

aioftp/client.py
```diff
@@ -107,36 +111,38 @@ async def __aexit__(self, exc_type, exc, tb):
...
+                 connection_timeout: Union[float, int, None] = None,
+                 read_speed_limit: Optional[int] = None,
+                 write_speed_limit: Optional[int] = None,
+                 path_timeout: Union[float, int, None] = None,
+                 path_io_factory: Type[pathio.PathIO] = pathio.PathIO,
+                 encoding: str = 'utf-8',
                  ssl=None,
                  parse_list_line_custom=None,
                  parse_list_line_custom_first=True,
                  passive_commands=("epsv", "pasv"),
```
pohmelie, line 124: Why types are declared just for subset of arguments?

pohmelie (changes requested):

pohmelie:
If we implement types, then we need mypy on CI of course.

.github/workflows/ci.yml
honglei: I'm newbie to codecov-action, 
In order to pass the tests,like  https://github.com/honglei/aioftp/actions/runs/5453748546 
I followed : https://github.com/codecov/codecov-action/blob/main/README.md

aioftp/client.py
honglei: Yes

aioftp/client.py
honglei: `x: int | float | None = None`  was first added to Python3.10 ,https://docs.python.org/3/whatsnew/3.10.html 
 [PEP 604](https://peps.python.org/pep-0604/), Allow writing union types as X | Y
So,  we'll target 3.10+ ?

aioftp/client.py
honglei: OK ,I'll double check later.

aioftp/client.py
honglei: https://pythonspeed.com/articles/stop-using-python-3.7/ says:
As of June 2023, [about 30% of packages downloaded from PyPI were for Python 3.7](https://pypistats.org/packages/__all__).

.github/workflows/ci.yml
pohmelie: I think `codecov` updates should be excluded in this PR and reimplemented is separate one if someone will need this at all.

aioftp/client.py
pohmelie: It is okay for now. Python release cycle has changed, and 3.7 has EOL a week ago (https://endoflife.date/python). Also, there are no current plans to add much features to project. So, I see no reason to not to target 3.10 or 3.11.

aioftp/client.py
```diff
@@ -33,10 +36,10 @@
...
 
 __all__ = (
-    "BaseClient",
-    "Client",
-    "DataConnectionThrottleStreamIO",
-    "Code",
+    'BaseClient',
+    'Client',
+    'DataConnectionThrottleStreamIO',
+    'Code',
```
pohmelie, line 42: Why quotes changed?

honglei (commented):
auto-formated by `black`, 
add missing type hints,
add ftp command `SIZE` in tests.

.github/workflows/ci.yml
pohmelie: Coverage working right now as is https://camo.githubusercontent.com/8631edb27e25ba8d58ab1df19fafe2f57991ce73cbe0098f3789f25270feba22/68747470733a2f2f636f6465636f762e696f2f67682f61696f2d6c6962732f61696f6674702f6272616e63682f6d61737465722f67726170682f62616467652e737667

honglei: https://app.codecov.io/gh/honglei/aioftp/commits

honglei:
I prefer pyftpdlib-style naming for FTP command process hander:
 `type` -> `ftp_TYPE`
 `list` ->  	 `ftp_LIST`

1) `type` and `list` are python key-words
2) these functions are used internal in class `Server`, so the change may not affect lib users.
3) the prefix `ftp_` is more clear than current names.

.github/workflows/ci.yml
pohmelie: Hm, https://app.codecov.io/gh/aio-libs/aioftp

pohmelie: I think, there are no pipelines for this PR until I approve it. https://app.codecov.io/gh/aio-libs/aioftp/pull/165

pohmelie:
> I prefer pyftpdlib-style naming for FTP command process hander: `type` -> `ftp_TYPE` `list` -> `ftp_LIST`
> 
>     1. `type` and `list` are python key-words
> 
>     2. these functions are used internal in class `Server`, so the change may not affect lib users.
> 
>     3. the prefix `ftp_` is more clear than current names.

It's okay to have a `list` method, since method does not shadow original `list`.

image

honglei:
> If we implement types, then we need mypy on CI of course.
There is no easy way to achieve this. like the following problems:
https://discuss.python.org/t/improve-typing-with-to-force-not-none-value/7840/3

pohmelie:
>There is no easy way to achieve this. like the following problems:
>https://discuss.python.org/t/improve-typing-with-to-force-not-none-value/7840/3
Just add check before using result:
```python
t: T | None = f()
if t is None:
    ...
t.bar()
```

honglei:
> > There is no easy way to achieve this. like the following problems:
> > https://discuss.python.org/t/improve-typing-with-to-force-not-none-value/7840/3
> > Just add check before using result:
> 
> ```python
> t: T | None = f()
> if t is None:
>     ...
> t.bar()
> ```

@pohmelie thanks，justlearned from https://mypy.readthedocs.io/en/stable/type_narrowing.html

honglei:
mypy and pytest finally all passed.

pohmelie (changes requested):
Need to fix all mypy errors. I expand linters with ruff (instead of flake8) and black formatter. And targeted python version 3.11.

honglei:
use python keyworkd list as function name in pyathio.py/PathIO caused some problems:
 
[    async def _open(self, path: pathlib.Path, *args: List[Any], **kwargs: dict[Any, Any]) -> BufferedRandom:](https://github.com/honglei/aioftp/blob/0cc53ff27276c5b241614fa7d86c0b6355859b6d/aioftp/pathio.py#L446)

ruff:
```
aioftp/pathio.py:446:54: UP006 Use `list` instead of `List` for type annotation
```
Changed to list:
```py3
async def _open(self, path: pathlib.Path, *args: list[Any], **kwargs: dict[Any, Any]) -> BufferedRandom:
```
mypy:
```
pathio.py:446:54: error: Function "aioftp.pathio.PathIO.list" is not valid as a
type  [valid-type]
    ...async def _open(self, path: pathlib.Path, *args: list[Any], **kwargs: ...
                                                        ^
pathio.py:446:54: note: Perhaps you need "Callable[...]" or a callback protocol?
pathio.py:447:16: error: No overload variant of "open" of "Path" matches
argument types "tuple[list?[Any], ...]", "dict[str, dict[Any, Any]]" 
[call-overload]
```

pohmelie:
> pathio.py:446

Just add type alias for this, something like `ListType = list` at top of the file.

honglei:
> ListType = list

Good idea, but bad ideasof reusing the name of build-in function name `list`.   https://stackoverflow.com/questions/9109333/is-it-bad-practice-to-use-a-built-in-function-name-as-an-attribute-or-method-ide

pohmelie:
> > ListType = list
> 
> Good idea, but bad ideas of reusing the name of build-in function name `list`. https://stackoverflow.com/questions/9109333/is-it-bad-practice-to-use-a-built-in-function-name-as-an-attribute-or-method-ide

It won't be changed. `pathio.list()` will remain `pathio.list()` it doesn't shadow original `list`. So lets make a workaround here, since it just type checks, not a real features/improvements or something. Think about it as `what API I want to use`, I think it is obvious, that `pathio.list()` is what client expect to see.

honglei:
you are right!
https://docs.python.org/3/library/ftplib.html#ftplib.FTP.dir

Nikita Melentev ***@***.***> 于 2023年8月14日周一 下午11:49写道：

PR #166, opened 2023-07-11
Don't output . and .. in client.list(*, recursive=True) as these create symlink loop on all ftp servers with UNIX systems

CrafterKolyan:
## What do these changes do?

Fix a bug when you use `client.list(path, recursive=True)` and never get out of it due to symlink loop with `.` and `..`.

Yes, if users had files `.` and `..` previously in the output of `client.list(path)` then they won't see them now.
Also `client.list(path, recursive=True)` will now work for UNIX systems

## Testing

Tested these changes on FTP server with UNIX filesystem

## Checklist

### DIFF
--- a/aioftp/client.py
+++ b/aioftp/client.py
@@ -781,25 +781,30 @@ class Client(BaseClient):
                 cls.directories = collections.deque()
                 return cls
 
-            async def __anext__(cls):
+              async def __anext__(cls):
                 if cls.stream is None:
                     cls.stream = await cls._new_stream(path)
                 while True:
-                    line = await cls.stream.readline()
-                    while not line:
-                        await cls.stream.finish()
-                        if cls.directories:
-                            current_path, info = cls.directories.popleft()
-                            cls.stream = await cls._new_stream(current_path)
-                            line = await cls.stream.readline()
-                        else:
-                            raise StopAsyncIteration
+                    continue_reading = True
+                    while continue_reading:
+                        line = await cls.stream.readline()
+                        while not line:
+                            await cls.stream.finish()
+                            if cls.directories:
+                                current_path, info = cls.directories.popleft()
+                                cls.stream = await cls._new_stream(current_path)
+                                line = await cls.stream.readline()
+                            else:
+                                raise StopAsyncIteration
 
-                    name, info = cls.parse_line(line)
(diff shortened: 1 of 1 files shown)


aioftp/client.py
```diff
@@ -781,26 +781,31 @@ def __aiter__(cls):
                 cls.directories = collections.deque()
                 return cls
 
-            async def __anext__(cls):
+              async def __anext__(cls):
```
CrafterKolyan, line 784:
```suggestion
            async def __anext__(cls):
```

aioftp/client.py
```diff
@@ -785,22 +785,27 @@ async def __anext__(cls):
...
+
+                        name, info = cls.parse_line(line)
+                        # skipping . and .. as these are symlinks in Unix 
+                        continue_reading = str(name) in (".", "..")
                     stat = cls.path / name, info
                     if info["type"] == "dir" and recursive:
                         cls.directories.append(stat)
                     return stat
 
+
```
CrafterKolyan, line 808:
```suggestion
```

history.rst
```diff
@@ -1,5 +1,6 @@
-x.x.x (xx-xx-xxxx)
+0.22.0 (xx-xx-xxxx)
```
pohmelie, line 1:
```suggestion
x.x.x (xx-xx-xxxx)

0.22.0 (xx-xx-xxxx)
```

aioftp/client.py
```diff
@@ -785,17 +785,21 @@ async def __anext__(cls):
...
+                            if cls.directories:
+                                current_path, info = cls.directories.popleft()
+                                cls.stream = await cls._new_stream(current_path)
+                                line = await cls.stream.readline()
+                            else:
+                                raise StopAsyncIteration
+
+                        name, info = cls.parse_line(line)
+                        # skipping . and .. as these are symlinks in Unix 
+                        continue_reading = str(name) in (".", "..")
```
pohmelie, line 802: It is simpler not to do nested loop. You can just add check.
```suggestion
                    if str(name) in (".", ".."):
                        continue
```

pohmelie (changes requested):
Fix linter error

CrafterKolyan:
Please rerun workflow

PR #169, opened 2023-11-26
Include symlink destination in path info.

rcfox:
## What do these changes do?

The existing `parse_list_line_unix` function goes to all the work of parsing symlink lines, but omits the actual destination of the symlink. This tiny change adds the parsed destination to the path's info dictionary.

One extra key is included in the `info` dict when issuing a `LIST` command to a unix server.

## Checklist

- [ ] I think the code is well written


pohmelie:
FTP protocol have nothing about `link` type. It is just internal part of current implementation. If you need all this information, you should use `parse_list_line_custom` and use your own parse code or copy-paste existing one with your fixes.

nE0sIghT:
+1 for this change.  
As of now the `size` reported by the `stat` is a link file size, but not a link target size.  

With current implementation there is no way to know if the size is valid.

PR #177, opened 2024-05-14
server can override and add custom command mappings

rupurt:
## What do these changes do?

The `Server` class can override and add custom command mappings. I'm submitting this as a draft PR without unit tests to open the conversation around if this is a desirable change that will be accepted.

No

## Related issue number

None

## Checklist

### DIFF
--- a/src/aioftp/server.py
+++ b/src/aioftp/server.py
@@ -660,6 +660,7 @@ class Server:
         data_ports=None,
         encoding="utf-8",
         ssl=None,
+        extra_commands_mapping={},
     ):
         self.block_size = block_size
         self.socket_timeout = socket_timeout
@@ -718,7 +719,7 @@ class Server:
             "syst": self.syst,
             "type": self.type,
             "user": self.user,
-        }
+        } | extra_commands_mapping
 
     async def start(self, host=None, port=0, **kwargs):
         """


rupurt:
@pohmelie do you have any suggestions to get this merged?

pohmelie:
> @pohmelie do you have any suggestions to get this merged?

You need to move this from draft. "Draft" mode implies you are not ready for code review.

image

rupurt:
Done! :beers:

setup.py
```diff
@@ -0,0 +1,4 @@
+from setuptools import setup
+
+if __name__ == "__main__":
+    setup()
```
pohmelie, line 4: Actually we do not need setupy.py, because of pyproject.toml file

src/aioftp/server.py
```diff
@@ -660,6 +660,7 @@ def __init__(
         data_ports=None,
         encoding="utf-8",
         ssl=None,
+        extra_commands_mapping={},
```
pohmelie, line 663: This mapping should be added to documentation

pohmelie (changes requested):

setup.py
rupurt: I originally thought I needed it for Nix but I found an override. I'll remove the commit

PR #181, opened 2024-12-19
add new datetime parsing format for client to support windows FTP server list output

ghost:
### Description:-
This PR enhances the **parse_list_line_windows** method to support multiple date/time formats when parsing Microsoft Windows dir command output. Previously, only one specific date/time format was supported. The following improvements were made:

**Changes:**

- Added support for two date/time formats:
"%m/%d/%Y %I:%M %p" (e.g., 12/18/2024 04:40 PM)
"%m-%d-%y %I:%M%p" (e.g., 12-18-24 04:40PM)

- Introduced a loop to try each format and parse the date/time correctly.

- Raised a ValueError if none of the formats match, ensuring robust error handling.

**Why:**
Windows FTP servers may use different date/time formats depending on system configurations and locale settings. Supporting multiple formats improves compatibility and reduces parsing failures.

image (9)

### DIFF
--- a/src/aioftp/client.py
+++ b/src/aioftp/client.py
@@ -518,7 +518,6 @@ class BaseClient:
         info = {}
         info["modify"] = self.format_date_time(date_time)
         next_space = line.index(" ")
-
         if line.startswith("<DIR>"):
             info["type"] = "dir"
         else:


src/aioftp/client.py
```diff
@@ -501,9 +501,19 @@ def parse_list_line_windows(self, b):
...
         date_time_str = " ".join([x for x in date_time_str if len(x) > 0])
         line = line[date_time_end + 1 :].lstrip()
+
+        possible_formats = ["%m/%d/%Y %I:%M %p", "%m-%d-%y %I:%M%p"]
+        date_time = None
         with setlocale("C"):
-            strptime = datetime.datetime.strptime
-            date_time = strptime(date_time_str, "%m/%d/%Y %I:%M %p")
+            for fmt in possible_formats:
+                try:
```
pohmelie, line 509: Use contextlib.suppress if you don't need to handle exception, but want to ignore it

src/aioftp/client.py
```diff
@@ -501,9 +501,19 @@ def parse_list_line_windows(self, b):
...
+                    date_time = datetime.datetime.strptime(date_time_str, fmt)
+                    break
+                except ValueError:
+                    continue
+            if not date_time:
```
pohmelie, line 514: Use "else" branch for "for" loop instead of value inspection

src/aioftp/client.py
```diff
@@ -501,9 +501,19 @@ def parse_list_line_windows(self, b):
         date_time_str = line[: date_time_end + 1].strip().split(" ")
         date_time_str = " ".join([x for x in date_time_str if len(x) > 0])
         line = line[date_time_end + 1 :].lstrip()
+
+        possible_formats = ["%m/%d/%Y %I:%M %p", "%m-%d-%y %I:%M%p"]
```
pohmelie, line 505: I think this should bu multiline, so new rules can be added without commit history loss

pohmelie (changes requested):

fti-akumari:
Sorry for the late response. I have addressed all the comments and made the necessary updates.

src/aioftp/client.py
```diff
@@ -501,9 +501,19 @@ def parse_list_line_windows(self, b):
         date_time_str = line[: date_time_end + 1].strip().split(" ")
         date_time_str = " ".join([x for x in date_time_str if len(x) > 0])
         line = line[date_time_end + 1 :].lstrip()
-        with setlocale("C"):
```
pohmelie, line 504: You remove setlocale, by the way. Need to return it back

pohmelie (changes requested):

src/aioftp/client.py
```diff
@@ -501,9 +501,19 @@ def parse_list_line_windows(self, b):
...
-            strptime = datetime.datetime.strptime
-            date_time = strptime(date_time_str, "%m/%d/%Y %I:%M %p")
+        
+        possible_formats = [
+        "%m/%d/%Y %I:%M %p",
```
pohmelie, line 506: Make sure linters are ok. This is not correct indentation.

src/aioftp/client.py
```diff
@@ -501,12 +501,24 @@ def parse_list_line_windows(self, b):
...
+                with contextlib.suppress(ValueError):
+                    date_time = datetime.datetime.strptime(date_time_str, fmt)
+                    break
+            else:
+                raise ValueError(f"Date/time format not recognized: {date_time_str}")
+
         info = {}
         info["modify"] = self.format_date_time(date_time)
         next_space = line.index(" ")
+
```
pohmelie, line 521: Why we need this extra empty line here?

pohmelie:
lint check fails, fix your code according linters, use `pre-commit run -a` before commit your code

src/aioftp/client.py
```diff
@@ -520,6 +531,7 @@ def parse_list_line_windows(self, b):
         filename = line[next_space:].lstrip()
         if filename == "." or filename == "..":
             raise ValueError
+
```
pohmelie, line 534: Why we need this additional empty line here?

PR #187, opened 2025-05-07
resolve tls issues (fixes #186)

### LINKED ISSUE #186: upgrade_to_tls does not work correctly
Kind of follow up to https://github.com/aio-libs/aioftp/issues/184. Explicit FTPS works now, however:

- Not possible to use the context API, since there is no way to call updgrade_to_tls() in between client.connect() and client.login(). 
- I was not able to download from pyftpdlib until I used this construction:
```                  
await client.upgrade_to_tls(self.ssl_context)
client._upgraded_to_tls = True # pylint: disable=protected-access
client._open_connection = functools.partial(asyncio.open_connection, ssl=client.ssl) # pylint: disable=protected-access
```
AFAIK, _upgraded_to_tls is never set in the actual code, and _open_connection is incorrect after update since it is not update with new ssl instance.

pohmelie:


pohmelie:
@astrand @sammichaels @bellini666 take a look and test on your servers please

astrand:
Branch works - our tests now passes without this:
```
                    client._upgraded_to_tls = True # pylint: disable=protected-access
                    client._open_connection = functools.partial(asyncio.open_connection, ssl=client.ssl) # pylint: disable=protected-access
```

sammichaels:
@pohmelie This is failing on my test.  The SSL handshake is timing out when opening the data connection.  I'll take a closer look as soon as I can.

pohmelie:
@sammichaels did you found something?

PR #188, opened 2025-05-21
Type Annotations

### LINKED ISSUE #123: No release notes since 0.13
This library has no release notes since version 0.13. This makes tracking the changes incredibly difficult, which can cause serious problems, especially since it seems like minor versions may include breaking changes (such as moving from 0.16.1 to 0.17).

### LINKED ISSUE #164: Add type hints to aioftp
Since the aioftp requires Python3.7+ right now, type hints may benefit the project.

sakosha:
## What do these changes do?

Adding type annotations to user facing API, internals and bug-fixes discovered during type checking.

`typing-extensions` are now required dependency before py3.11.

users extending Client should replace `self.stream` w/ `self._stream` to preserve old behavior.

## Related issue number

Fixes #164

## Checklist

These changes are large. I hope it won't be hard to review


pohmelie:
`pre-commit` linters-set failed, run `pre-commit run -a` until all green and add changes to this pr. Also, I'm not sure this will work for python 3.9 in terms of modern type hinting.

pohmelie:
CI fails, fix please

> I don't quite get what you suggesting about pytohn3.9

I was afraid of type hints won't work for python 3.9, but it looks like they are ok

pohmelie:
CI mypy fails @sakosha

sakosha:
Yep. I'll try troubleshooting it on Th.

pohmelie (changes requested):

pohmelie:
Good job @sakosha. I reviewed client and some other files. Did not look at server and pathio files, but lets start from client

src/aioftp/client.py
```diff
@@ -500,10 +615,14 @@ def parse_list_line_windows(self, b):
...
-        line = b.decode(encoding=self.encoding).rstrip("\r\n")
+        if isinstance(b, bytes):
+            line = b.decode(encoding=self.encoding).rstrip("\r\n")
+        else:
+            line = b.rstrip("\r\n")
+
         date_time_end = line.index("M")
-        date_time_str = line[: date_time_end + 1].strip().split(" ")
-        date_time_str = " ".join([x for x in date_time_str if len(x) > 0])
+        date_time_str_list = line[: date_time_end + 1].strip().split(" ")
```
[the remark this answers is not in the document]
sakosha: mypy cannot disambiguate `str` & `Sequence[str]`, and split() returns list[str]

and before we had a single name for 2 of these, and it is not allowed before mypy 2.0

src/aioftp/client.py
```diff
@@ -261,8 +368,10 @@ async def command(
             when logging
         :type censor_after: :py:class:`None` or :py:class:`int`
         """
-        expected_codes = wrap_with_container(expected_codes)
-        wait_codes = wrap_with_container(wait_codes)
+        expected_code_strs = wrap_with_container(expected_codes)
+        expected_codes = tuple([Code(x) for x in expected_code_strs])
+        wait_code_strs = wrap_with_container(wait_codes)
+        wait_codes = tuple([Code(x) for x in wait_code_strs])
```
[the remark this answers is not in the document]
sakosha: maybe we can refactor it to `wrap_into_codes` in common.py to wrap `Sequence[str]` into `Sequence[Code]`, and remove/inline other usages of wrap_with_container?

src/aioftp/client.py
```diff
@@ -444,7 +556,10 @@ def parse_list_line_unix(self, b):
         :return: (path, info)
         :rtype: (:py:class:`pathlib.PurePosixPath`, :py:class:`dict`)
         """
-        s = b.decode(encoding=self.encoding).rstrip()
+        if isinstance(b, bytes):
+            s = b.decode(encoding=self.encoding).rstrip()
+        else:
+            s = b
```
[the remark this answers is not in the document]
sakosha: I used documentation comments at the start, and I assumed it was allowed.

should we also remove this check from `parse_mlsx_line`?

sakosha:
Thanks for the fast review! I'll wait for further feedback.

pohmelie:
> hello! I believe TypedDict issue is the only one left. I think we should left it as is, to not break our users' client code.

I need to review other files. Sorry for a delay, occupied at work hardly. Will try to review this on this week

.github/workflows/ci-cd.yml
```diff
@@ -16,6 +16,17 @@ jobs:
...
+  type-check:
+    needs: lint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ env.lowest_python_version }}
+    - run: python -m pip install -e ./[dev]
+    - run: mypy
```
pohmelie, line 28: Why we need separate step for mypy? Since it is in pre-commit step already

.pre-commit-config.yaml
```diff
@@ -48,3 +48,10 @@ repos:
     name: Check with ruff
     entry: ruff
     args: ["check", "--fix", "."]
+
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.15.0
+  hooks:
+  - <<: *python-linters
+    id: mypy
+    name: Type check with mypy
```
pohmelie, line 57: This should be under `repo: local` part as ruff did

pyproject.toml
```diff
@@ -40,6 +40,10 @@ classifiers = [
     "Topic :: Internet :: File Transfer Protocol (FTP)",
 ]
 
+dependencies = [
+    "typing-extensions>=4.13.2",  # 3.9 typing support
```
pohmelie, line 44: I think we need a TODO here: remove typing extension after 3.9 support drop. Am I right?

src/aioftp/common.py
```diff
@@ -40,24 +49,58 @@
...
 
-def _now():
+def _now() -> float:
     return asyncio.get_running_loop().time()
 
 
-def _with_timeout(name):
-    def decorator(f):
+_WithTimeOutP = ParamSpec("_WithTimeOutP")
+_WithTimeOutParamR = TypeVar("_WithTimeOutParamR")
```
pohmelie, line 57: What means end `R` in naming?

src/aioftp/common.py
```diff
@@ -125,17 +168,21 @@ class AsyncListerMixin:
...
             items.append(item)
         return items
 
-    def __await__(self):
+    def __await__(self) -> Generator[None, None, list[T]]:
         return self._to_list().__await__()
 
+    @abc.abstractmethod
+    def __aiter__(self) -> AsyncIterator[T]:
+        pass
```
pohmelie, line 182: We don't need `pass` in abstract methods. Common way is to add a docstring instead of abstract method body, since abstract method body will never be evaluated:
```py
@abstractmethod
def foo(self) -> T:
    """
    This method do foo
    """
```

src/aioftp/pathio.py
```diff
@@ -1,21 +1,43 @@
...
     AbstractAsyncLister,
     AsyncStreamIterator,
     with_timeout,
 )
+from .errors import PathIOError
+
+if TYPE_CHECKING:
+    from _typeshed import OpenBinaryMode, ReadableBuffer
+
+    from .server import Connection
```
pohmelie, line 39: If this `if TYPE_CHECKING` is about circular imports, than lets resolve this via moving Connection to common module

src/aioftp/pathio.py
```diff
@@ -90,41 +127,94 @@ async def wrapper(*args, **kwargs):
...
+
+class PathIONursery(Generic[PathIOT]):
+    def __init__(self, factory: type["PathIOT"]) -> None:
+        self.factory = factory
+        self.state: Union[list[Node], None] = None
+
+    def __call__(
+        self,
+        timeout: Union[float, int, None] = None,
+        connection: Union["Connection", None] = None,
```
pohmelie, line 162: Fix this "Connection" to Connection after moving Connection to common

src/aioftp/pathio.py
```diff
@@ -90,41 +127,94 @@ async def wrapper(*args, **kwargs):
...
+        state: Union[list["Node"], None] = None,
+    ) -> PathIOT:
+        instance = self.factory(timeout=timeout, connection=connection, state=self.state)
         if self.state is None:
             self.state = instance.state
         return instance
 
 
-def defend_file_methods(coro):
+# DefendFileMethodsSelf = TypeVar("DefendFileMethodsSelf")
```
pohmelie, line 171: If it is obsolete, then lets remove this

src/aioftp/pathio.py
```diff
@@ -90,41 +127,94 @@ async def wrapper(*args, **kwargs):
...
-class AbstractPathIO(abc.ABC):
+class PathlibOpenKwargs(TypedDict, total=False):
+    mode: "OpenBinaryMode"
+    buffering: int
+    encoding: Union[str, None]
+    errors: Union[str, None]
+    newline: Union[str, None]
+
+
+PathT = TypeVar("PathT", bound=PurePath)
```
pohmelie, line 214: Hm, why we need an extra PathType here?

src/aioftp/pathio.py
```diff
@@ -1,21 +1,43 @@
...
 from .common import (
     DEFAULT_BLOCK_SIZE,
     AbstractAsyncLister,
     AsyncStreamIterator,
     with_timeout,
 )
+from .errors import PathIOError
+
+if TYPE_CHECKING:
+    from _typeshed import OpenBinaryMode, ReadableBuffer
```
pohmelie, line 37: Hm, never seen this before. Can you explain why we need this?

src/aioftp/pathio.py
```diff
@@ -716,31 +899,39 @@ async def rmdir(self, path):
             raise OSError("Directory not empty")
 
         parent = self.get_node(path.parent)
-        for i, node in enumerate(parent.content):
-            if node.name == path.name:
+        if parent is None:
+            raise FileNotFoundError
+        if not isinstance(parent.content, list):
+            raise NotADirectoryError
+        for i, n in enumerate(parent.content):
```
pohmelie, line 906: No need to shortify names, use full name "node" here, because of readability

pohmelie:
@sakosha
- Please replace suffixes for types like P, R, T to fully named ParamSpec, TypeVar, Type for readability.
- Pull changes from master
- I will review server part after and we can finally merge this big work to master 🎉

src/aioftp/common.py
sakosha: `R` means return TypeVar.

see:
https://peps.python.org/pep-0612/
basically is one has to type hint a simple decorator they declare a ParamSpec & TypeVar like this
```python
from typing import ParamSpec, TypeVar

P = ParamSpec('P')
R = TypeVar('R')
```

but I think it's a typo on my side as they should be:
```python
_WithTimeOutP = ParamSpec("_WithTimeOutP")
_WithTimeOutR = TypeVar("_WithTimeOutR")
```

src/aioftp/pathio.py
sakosha: oops. my bad :D

pyproject.toml
pohmelie: Ok, add comment here then

src/aioftp/common.py
pohmelie: Yep, but in pep it is `R = TypeVar("R")`, so it is obvious and R is just an example. In case of named type, we should use this exact names `_WithTimeOutParamSpec` and `_WithTimeOutTypeVar`

src/aioftp/pathio.py
pohmelie: Ok, lets just rename it to `PathType` then

src/aioftp/common.py
sakosha: Let's use `...ReturnType` instead as it isn't a concrete type anyway, so it should look distinct.

sakosha:
Hi again! I believe I fixed all of the comments.

sakosha:
Also, I think we should ping others to see if we have any issues in fork.
Of course after we done w/ code review.

src/aioftp/common.py
```diff
@@ -164,24 +223,45 @@ class AbstractAsyncLister(AsyncListerMixin, abc.ABC):
...
+
+
+AsyncEnterableReturnType = TypeVar("AsyncEnterableReturnType", bound=AsyncContextManager, covariant=True)
+
+
+class AsyncEnterableInstanceProtocol(Protocol[AsyncEnterableReturnType]):
+    async def __aenter__(self) -> AsyncEnterableReturnType: ...
+
+    async def __aexit__(self, *args: Any, **kwargs: Any) -> None: ...
+
```
pohmelie, line 255: This empty lines between one-liners should be removed, I think

src/aioftp/pathio.py
```diff
@@ -482,217 +617,274 @@ class AsyncPathIO(AbstractPathIO):
     :type executor: :py:class:`concurrent.futures.Executor`
     """
 
-    def __init__(self, *args, executor=None, **kwargs):
-        super().__init__(*args, **kwargs)
+    executor: Union[Executor, None]
```
pohmelie, line 620: Hm, do we need this here, on class level? Is `self.executor: Union[Executor, None] = executor` from `__init__` not enough?

src/aioftp/server.py
```diff
@@ -36,11 +48,8 @@
     "Server",
 )
 
-IS_PY37_PLUS = sys.version_info[:2] >= (3, 7)
-if IS_PY37_PLUS:
-    get_current_task = asyncio.current_task
-else:
-    get_current_task = asyncio.Task.current_task
+
+get_current_task = asyncio.current_task
```
pohmelie, line 52: Since we have no dispatch, then we do not need this special binding. Just replace `get_current_task` with `asyncio.current_task` everywhere

src/aioftp/server.py
```diff
@@ -375,16 +325,20 @@ def acquire(self):
...
         Release, incrementing the internal counter by one.
         """
-        if self.value is not None:
+        if self.value is not None and self.maximum_value is not None:
             self.value += 1
             if self.value > self.maximum_value:
                 raise ValueError("Too many releases")
 
 
+ConnectionConditionsP = ParamSpec("ConnectionConditionsP")
```
pohmelie, line 338: P -> ParamSpec, R -> ReturnType. And all places below too

src/aioftp/server.py
```diff
@@ -1063,9 +1122,6 @@ async def user(self, connection, rest):
             connection.user = user
         elif state == AbstractUserManager.GetUserResponse.ERROR:
             code = "530"
-        else:
-            message = f"Unknown response {state}"
-            raise NotImplementedError(message)
```
pohmelie, line 1068: So, this removed because of `get_user` can't return anything else since typing description?

pohmelie (changes requested):

src/aioftp/server.py
sakosha: Yes, state should be a member of `GetUserResponse` enum. We can leave these lines, but this code is unreachable.

PR #193, opened 2025-06-27
fix #192

### LINKED ISSUE #123: No release notes since 0.13

### LINKED ISSUE #192: Incorrect type annotation
The following code gives error while type checking:
```python
client = aioftp.Client()
stream = await client.download_stream(...)
```
mypy error:
```
error: Incompatible types in "await" (actual type "DataConnectionThrottleStreamIO", expected type "Awaitable[Any]")  [misc]
```
I guess this is caused by `async_enterable` decorator wrapping `Client.get_stream` method.

sakosha:
Fixes #192

## Checklist

### DIFF
--- a/src/aioftp/client.py
+++ b/src/aioftp/client.py
@@ -1229,7 +1229,7 @@ class Client(BaseClient):
         source: PathLike,
         *,
         offset: int = 0,
-    ) -> DataConnectionThrottleStreamIO:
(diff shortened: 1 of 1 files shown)


sakosha:
not sure why codecov fails.

PR #197, opened 2025-08-18
fix year assumption in client list command

### LINKED ISSUE #196: Client list using "LIST" parses dates more than 6 months in the past without a year as the next year
I have a server that has files from January - August of this year. The files from January and February have no year (since it is assumed to be this year) and they are being read as last modified in 2026. The other files more recent read the correct year and the device has the correct time on it.

anaboccanfuso:
## What do these changes do?

Fixes an issue in the `parse_ls_date function` where dates more than half a year in the past were incorrectly showing up with next year's date.

The original code was using `HALF_OF_YEAR_IN_SECONDS` to determine if a date should be adjusted to the next year, which caused January dates to incorrectly show up as 2026 instead of 2025 when the current date is in August 2025.

The fix:
1. Removes the dependency on `HALF_OF_YEAR_IN_SECONDS`
2. Changes the date parsing logic to use the current year by default
3. Only adjusts the year to the previous year if the resulting date would be 
   in the future

This ensures dates are correctly interpreted regardless of how far in the past they are, fixing timestamp issues in LIST command responses.

No.

## Related issue number
Fixes #196

## Checklist

### DIFF
--- a/src/aioftp/client.py
+++ b/src/aioftp/client.py
@@ -27,7 +27,6 @@ from .common import (
     DEFAULT_PORT,
     DEFAULT_USER,
     END_OF_LINE,
-    HALF_OF_YEAR_IN_SECONDS,
     TWO_YEARS_IN_SECONDS,
     AbstractAsyncLister,
     AsyncEnterableInstanceProtocol,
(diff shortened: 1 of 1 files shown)


Vizonex (changes requested):
Looks good but may I recommend adding a test like @sakosha has said?

PR #207, opened 2026-08-03
feat: read welcome message from a variable

rwakulszowa:
## What do these changes do?

Allow setting a welcome message (aka banner) through a class variable. This is the message that the server replies with on the first connection.

Yes, but it's just a hardcoded string.

## Checklist

### DIFF
--- a/src/aioftp/server.py
+++ b/src/aioftp/server.py
@@ -697,6 +697,7 @@ class Server:
         data_ports: Iterable[int] | None = None,
         encoding: str = "utf-8",
         ssl: ssl.SSLContext | None = None,
+        welcome_message: str = "welcome",
     ) -> None:
         self.block_size = block_size
         self.socket_timeout = socket_timeout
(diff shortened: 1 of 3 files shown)


rwakulszowa:
Just a tiny thing that came up when I was using the library. It can be achieved by overriding the `greeting` method, but I figured I may not be the only person doing the exact same thing, so maybe it makes sense to have an easier way of doing this.

If you want, I can also add a CLI option.

For comparison, `pyftpdlib` does allow this:
https://github.com/giampaolo/pyftpdlib/blame/839ce0aa57d091873b56eaa657fca11e3dd23937/docs/cli.rst#L94

src/aioftp/server.py
```diff
@@ -678,6 +678,8 @@ class Server:
     :type ssl: :py:class:`ssl.SSLContext`
     """
 
+    welcome_message = "welcome"
```
pohmelie: I think it is better to pass `welcome_message` to `__init__`. Since there is no class-var mechanic used anywhere in codebase

pohmelie (changes requested):

pohmelie:
I will fix linter errors later today

pohmelie:
Thank you for the contribution

PR #208, opened 2026-08-10
feat: un-hardcode SYST response

rwakulszowa:
## What do these changes do?

Un-hardcode SYST response
It should identify the underlying OS. Allow the caller to specify the value.

The SYST response is no longer hardcoded to UNIX

## Checklist


pohmelie:
I think I will merge this, but what is the reason for this? Anyone uses SYST at all? I mean... the value of SYST response

rwakulszowa:
Whether anyone actually cares about the result in practice - I have no idea. The only reason I implemented this was because it's a well known (albeit not too useful) command and someone _might_ wanna use it, so I figured it's better to have it configurable.

PR #210, opened 2026-08-13
Use asyncio.timeout instead of async_timeout.timeout

billchenchina:


pohmelie:
We will drop python 3.10 in 2 months, so no reasons for this changes

billchenchina:
:)

PR #212, opened 2026-09-18
Apply connection_timeout to passive connections

Cycloctane:
## What do these changes do?

Client's connection_timeout option currently is only available when connecting control ports. This PR makes client also apply timeout settings when opening new passive (data) connections.

## Checklist


tests/test_connection.py
```diff
@@ -318,3 +318,18 @@ async def test_system_type(pair_factory, Server):
...
+
+    async def slow_open_connection(host, port, **kwargs):
+        await asyncio.sleep(1)
+        return await open_connection(host, port, **kwargs)
+
+    async with pair_factory(Client(connection_timeout=0.5), logged=True) as pair:
+        monkeypatch.setattr("asyncio.open_connection", slow_open_connection)
+        with pytest.raises(asyncio.TimeoutError):
+            await pair.client.get_passive_connection()
+        monkeypatch.undo()
```
pohmelie: Use context manager instead of `undo` call

pohmelie (changes requested):

